Issue #261, opened 2026-03-29
feat: add AI model info to ping MCP tool response


lwgray:
> **🎟️ PyCon 2026 Sprint Issue** — Reserved for the [PyCon 2026 sprint](https://github.com/lwgray/marcus/blob/develop/docs/sprints/pycon-2026.md) (May 2026 · Long Beach, CA). **At the sprint?** Comment below to claim it. **Working outside the sprint?** Please check in first — sprint participants get first pick.

---

## Background

**What is Marcus?**
Marcus is a system that coordinates multiple AI agents (Claude, GPT-4, etc.) to work on software projects in parallel. Think of it like a project manager: Marcus breaks a project into tasks, assigns them to AI workers, and tracks progress on a shared kanban board.

**What is the `ping` tool?**
The `ping` tool is the first MCP tool any agent calls to verify Marcus is alive. It returns basic status information about the running server — the "hello" response of the Marcus protocol.

---

## The Problem

`ping` tells you Marcus is running and which kanban backend is active, but it says nothing about which AI model Marcus is using. If you start Marcus with an environment variable override and then want to confirm which model is actually loaded, there is no quick way to check — you have to grep through log files.

**Relevant files:**
- `src/marcus_mcp/tools/system.py` — contains the `ping()` function (approximately line 72)

---

## Why it matters

A one-command sanity check for which AI model is running is essential during development and debugging. When users override the model via environment variable, they need a fast way to confirm the override took effect. Without this, debugging misconfiguration requires digging through logs.

---

## Acceptance Criteria

- [ ] The `ping` response includes the active AI provider name
- [ ] The `ping` response includes the active AI model name
- [ ] The `ping` response includes the configured temperature
- [ ] All existing `ping` fields are unchanged
- [ ] The fix is covered by at least one automated test

---

## Getting help

- Comment on this issue — maintainers check frequently during the sprint
- See the [sprint menu](https://github.com/lwgray/marcus/blob/develop/docs/sprints/pycon-2026.md) for project overview
- Ask in the sprint room

lwgray:
Related: #264 — persist AI config at startup so `./marcus status` reports accurate info (the CLI counterpart to this ping tool enhancement)

lwgray:
> **🎟️ PyCon 2026 Sprint Issue** — This issue is reserved for the [PyCon 2026 sprint](https://github.com/lwgray/marcus/blob/develop/docs/sprints/pycon-2026.md) (May 2026 · Long Beach, CA).
[...]

PR #262, opened 2026-03-29
feat: show AI model info in marcus status

lwgray:
## Summary

Two fixes for model-related visibility and behavior:

### 1. Show AI model info in `./marcus status`
Adds AI model, provider, and temperature to CLI status output so you can verify what model the running server loaded.

```
✅ Marcus is running
   PID: 12345
   Provider: sqlite
   AI Model: claude-haiku-4-5-20251001
   AI Provider: anthropic
   Temperature: 0.1
```

### 2. Fix domain discovery JSON parsing for Haiku 4.5
Haiku 4.5 wraps JSON responses in markdown fences, causing `json.loads()` to fail and falling back to generic "Project Domain" name. Replaced with `parse_ai_json_response()` which already handles fences/preamble (used everywhere else in the PRD parser).

## Context
- Debugging deprecated model (404 errors) was hard — no way to see what model was loaded
- "Design Project Domain" instead of "Design Snake Game Engine" was caused by domain discovery silently failing on every run with Haiku 4.5

## Related
- #261 (same AI info for MCP ping tool)

## Test plan
- [x] Pre-commit hooks pass
- [ ] `./marcus status` shows AI info
- [ ] Run experiment — domain names should be project-specific, not "Project Domain"

### DIFF
--- a/marcus
+++ b/marcus
@@ -365,9 +365,13 @@ class MarcusCLI:
         """Check Marcus status"""
         if self._is_running():
             pid = self._get_pid()
+            config = get_config()
             print("✅ Marcus is running")
             print(f"   PID: {pid}")
             print(f"   Provider: {get_kanban_provider()}")
+            print(f"   AI Model: {config.get('ai.model', 'unknown')}")
+            print(f"   AI Provider: {config.get('ai.provider', 'unknown')}")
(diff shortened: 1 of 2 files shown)


lwgray:
Follow-up: #264 — persist AI config at startup so status reads from runtime state instead of current config file (addresses the accepted limitation noted in this PR)

PR #317, opened 2026-04-07
fix: wire up resilience system and fix recovery for worktree agents

lwgray:
## Summary

- **Wire recovery handoff into task instructions** — when a lease expires and a task is recovered, the next agent now receives the previous agent's progress, branch name, and `git merge` instructions so they can pick up committed work
- **Fix gridlock false positives** — detection now uses task-state (all TODO blocked + zero in-progress) instead of request count, which was just measuring normal agent polling
- **Aggressive lease defaults** — 90s lease + 30s grace = ~2min recovery time instead of the previous 2.5 hours, which made recovery useless for experiments
- **Worktree-aware recovery** — instructions now tell agents to `git merge marcus/<dead-agent>` instead of `git log --author` which doesn't work across worktrees
- **Delete dead code** — removed TaskRecoveryManager (never instantiated, superseded by LeaseManager)
- **Document error_strategies.py** — marked as available but not integrated into MCP layer

## What was broken

The resilience system had all the pieces but wasn't wired end-to-end:
1. `RecoveryInfo` was created on lease expiry but never shown to the next agent
2. Gridlock detector fired on agent polling frequency (false positives every 90s)
3. Default 2-hour lease meant dead agents weren't detected for 2.5 hours
4. Recovery instructions referenced `git log --author=X` which shows nothing in a fresh worktree

## Test plan

- [x] 8 unit tests for recovery handoff layer (`test_recovery_handoff.py`)
- [x] 10 unit tests for gridlock detection accuracy (`test_gridlock_detector.py`)
- [x] 8 integration tests for full lease lifecycle (`test_resilience_e2e.py`)
- [x] 48 existing lease/timeout tests still pass
- [x] 19 existing workflow enforcement tests still pass
- [x] All pre-commit hooks pass (mypy, black, isort, flake8, bandit)
- [ ] Manual: run Marcus experiment, kill a tmux pane, verify recovery + reassignment


lwgray:
Thanks for the thorough review. Addressing the items:                                                                                                                 
                                                                                                                                                                        
  ✅ Done                                                                                                                                                               
                                                                                                                                                                        
  Test gap for report_task_progress lease recreation — Added TestKillAndPickup::test_report_task_progress_recreates_lease_after_recovery that invokes the real          
  report_task_progress function (not a copy of its logic) to verify the recreation branch fires. Confirmed via sed-test that removing the else branch at task.py:1813   
  causes this test to fail. 92 resilience tests now passing.                                                                                                            
                                                                                                                                                                        
  🤔 Pushing back on two items                                                                                                                                          
                                                                                                                                                                        
  1. Gridlock request_threshold (Codex feedback)                                                                                                                        
                                                                                                                                                                        
  I don't think we should add this back. The whole point of the rewrite was that request_threshold was the source of false positives — agents poll every ~30s, so 3     
  requests in 5 minutes is steady-state noise, not a signal.                                                                                                            
                                                                                                                                                                        
  The new conditions are strict enough on their own:                                                                                                                    
  - len(todo_tasks) > 0 — work exists                                                                                                                                   
  - len(blocked_tasks) == len(todo_tasks) — ALL of it is blocked                                                                                                        
  - len(in_progress_tasks) == 0 — nothing can complete to unblock anything                                                                                              
                                                                                                                                                                        
  When all three are true, forward progress is impossible regardless of how many agents notice. Adding distinct_agents >= 3 would just delay the alert without changing 
  the decision — and it would re-introduce the false positive class we fixed (single-agent or 2-agent projects in true deadlock that wouldn't alert until a third agent 
  happens to poll).                                                                                                                                                     
                                                                                                                                                                        
  The "low concurrency premature alert" scenario the review mentions is actually correct behavior: a 1-agent project with all TODO tasks blocked IS deadlocked. We want 
  to know.                                                                                                                                                              
                                                                                                                                                                        
  2. datetime.now() inside the lock                                                                                                                                     
                                                                                                                                                                        
  Capturing the timestamp before lock acquisition would be worse, not better:                                                                                           
  - datetime.now() is nanoseconds — the lock-hold overhead is negligible                                                                                                
  - Pre-lock capture gives a stale timestamp by however long we wait for the lock (could be milliseconds under contention)                                              
  - For a lease system, you want "now" to reflect when the lease is actually being created/checked, not when the function was called                                    
                                                                                                                                                                        
  The current placement is intentional and correct.                                                                                                                     
                                                                                                                                                                        
  Other items                                                                                                                                                           
                                                                                                                                                                        
  Marcus error framework for recovery exceptions — Fair point, but the except Exception at the end of recover_expired_lease is intentionally broad because we don't want
   a transient kanban hiccup to leave the system in a half-recovered state. I'd rather log and return False so the monitor retries on the next cycle. Adding            
  exc_info=True to the logger.error is a reasonable improvement though — happy to do that in a follow-up.                                                               
                                                                                                                                                                        
  Type safety (fractional hours) — Agreed it's awkward. The constructor signature default_lease_hours: float = 0.025 does need a comment to be readable. The dataclass  
  already has # 90 seconds — fast failure detection in marcus_config.py. A timedelta-based API would be cleaner but breaks backward compat. Worth tracking as a separate
   refactor.                                                                                                                                                            
                                                                                                                                                                        
  Per-task locking — Premature optimization for our scale. The current lease_lock only contends during create/renew/check operations, all of which are sub-millisecond. 
  We'd need profiling evidence before splitting it.                                                                                                                     
                                                                                                                                                                        
  Ready to merge if the gridlock and datetime points are acceptable. Happy to discuss further if you disagree on the gridlock reasoning.

Issue #320, opened 2026-04-10
feat: contract-first task decomposition (generate contracts before decomposing)


lwgray:
## Background

Supersedes #267. The original framing proposed a topology classifier
with a file-overlap matrix and a 6-topology decision tree; that did
not survive architectural review. This issue is the successor with
a smaller, concrete proposal: **move contract generation before
task decomposition, and have the decomposer produce tasks that each
own one side of a contract.**

Experiment 1 (snake game, hand-crafted TypeScript contract,
2026-04-10, `docs/audit-reports/snake-game-contract-based-v1`)
validated the core mechanism. Experiment 2 (Python log analyzer) is
pending.

## Problem

Marcus decomposes projects by product features before the design
phase runs. For tightly-coupled problems — where multiple features
naturally live in the same file — this produces tasks that collide
at the file level during implementation, and one agent ends up
writing the entire product while the other produces ghost code.

### Current evidence

- **v0.3.0 CHANGELOG**: "Known limitations: lopsided contributions
  (73/27 split)". Listed as a standing, unresolved finding.
- **Snake game audits (v106-run1, v108-run0, v108-run1,
  2026-03-30)**: three runs, three "Single-Author Product" verdicts.
  Feature-based decomposition put `Movement` and `Food Collection`
  into separate tasks, but both tasks required editing the same
  `gameLogic.ts` / `game.js` file. Whichever agent started first
  absorbed all the work. The contribution split on those runs was
  100/0, 98.5/1.5, and 96.9/3.1 respectively.
- **Snake game contract-based run (2026-04-10)**: two agents, a
  hand-crafted `types.ts` contract describing `GameEngine`,
  `GameState`, `Direction`, `Coord`, and tasks that each owned one
  side of the contract. Contribution split: ~30/70 by line count.
  Both agents produced substantive, non-ghost product code. Clean
  merge, zero file conflicts, contract unmodified. See
  `docs/audit-reports/snake-game-contract-based-v1`.

The new run is the first Marcus experiment on a tightly-coupled
problem that did not produce a Single-Author Product verdict.

## Root cause

`AdvancedPRDParser.parse_prd_to_tasks()` generates tasks from
functional requirements before any architectural analysis has
happened. Design autocomplete (#297) runs *after* decomposition,
against already-shaped tasks, so its output — including interface
contracts — has no input into how the tasks were carved.

The result: feature-nouns from the PRD drive task shapes, not
interface boundaries from the architecture. On problems where
[...]

Issue #304, opened 2026-04-03
perf: parallelize design autocomplete across multiple design tasks


lwgray:
## Problem

Design autocomplete (Phase A) processes design tasks sequentially. Each design
task requires 4 LLM calls (3 artifact documents + 1 decisions). For a standard
project with 3 design tasks, that's 12 sequential LLM calls. For enterprise
with 10 design tasks, 40 sequential calls.

Each LLM call takes ~30-60 seconds. A 3-domain project takes ~4-6 minutes
for design autocomplete. A 10-domain project would take ~20-30 minutes —
far beyond the 10-minute project creation timeout.

## Current Code

In _generate_design_content() (src/integrations/nlp_tools.py), the outer
loop over design tasks is sequential:

    for task in design_tasks:
        # 4 sequential LLM calls per task
        for spec in _DESIGN_ARTIFACT_SPECS:
            response = await llm.analyze(...)
        dec_response = await llm.analyze(...)

## Proposed Fix

Parallelize at two levels:

### Level 1: Across design tasks (asyncio.gather)
All design tasks are independent — they cover different domains. Process
them concurrently:

    results = await asyncio.gather(*[
        _generate_single_design(task, llm, ...)
        for task in design_tasks
    ])

### Level 2: Within each design task (optional)
The 3 artifact calls within a single task could also run in parallel since
they're independent documents. But this depends on LLM rate limits.

## Target

Whether 1 or 100 design tasks, the design process should complete within
the 10-minute project creation timeout. With parallelization across tasks,
the wall-clock time is dominated by the slowest single task (~2 minutes),
not the sum of all tasks.

## Acceptance Criteria

- [ ] Design tasks processed via asyncio.gather, not sequential loop
- [ ] 3-domain project completes design in ~2 minutes (not 6)
- [ ] 10-domain project completes design in ~2 minutes (not 30)
- [ ] Graceful degradation: if one task's LLM call fails, others still complete
- [ ] Rate limiting respected (if LLM provider has concurrent request limits)

## Related

- #297 — Design autocomplete (current sequential implementation)
- #250 — Git worktree isolation (depends on design completing before workers start)

lwgray:
Shipped in #319 (merged to develop at 17daaa6). Smoke test evidence: 10-task enterprise project went from ~25-33 min sequential to 145s parallel (~10-14× speedup, 27% of 10-min budget). Bonus: the run caught 15 transient Anthropic timeouts that @with_retry auto-recovered.

PR #322, opened 2026-04-11
refactor: domain-keyed contract generation + enable integration verification on experiments (#320 PR 1)

lwgray:
## Scope

This is **PR 1 of 3** for #320 (contract-first task decomposition). It is a **no-op for users** — the feature-based decomposition path is functionally unchanged. What it does is set up the plumbing PR 2 will need.

See [#320 for the full 3-PR plan and architectural rationale](https://github.com/lwgray/marcus/issues/320).

## Refactor: domain-keyed contract generation

The artifact/decisions LLM logic in \`_generate_design_content()\` was per-task. It's now per-domain, with a new public entry point that takes a \`{domain_name: description}\` dict and returns results keyed by domain name — no Task objects needed.

**Before:**
\`\`\`python
# Callers had to have Task objects to generate contracts
await _generate_design_content(tasks, ...)
\`\`\`

**After:**
\`\`\`python
# Task-free path — what PR 2's contract-first decomposer will call
await _generate_contracts_by_domain(
    domains={\"Authentication\": \"Design auth with JWT...\"},
    ...
)

# Task-based path — still works, now a thin adapter
await _generate_design_content(tasks, ...)  # byte-identical output
\`\`\`

Signature changes on the inner helpers:

| Helper | Before | After |
|---|---|---|
| \`_generate_single_artifact\` | takes \`task\` | takes \`(domain_name, domain_description)\` |
| \`_generate_single_decisions\` | takes \`task\` | takes \`(domain_name, domain_description)\` |
| \`_process_design_task\` → \`_process_design_domain\` | takes \`task\` | takes \`(domain_name, domain_description)\` |

\`_generate_design_content()\` is now a thin adapter:
1. Converts tasks to \`{domain_name: description}\` (strips \`\"Design \"\` prefix)
2. Delegates to \`_generate_contracts_by_domain()\`
3. Maps results back to task-keyed shape
4. Mutates task state atomically (\`status=DONE\`, \`assigned_to=\"Marcus\"\`, \`auto_completed\` label)

From the caller's perspective the behavior is **byte-identical**.

## Fix: remove \"experiment\" from integration verification skip list

\`IntegrationTaskGenerator.should_add_integration_task\` previously skipped projects whose description contained the keyword \`\"experiment\"\`. That silently disabled integration verification on every Marcus experiment run, which meant we could not measure whether integration verification catches the bugs that multi-agent coordination changes (like contract-first decomposition) introduce.

Experiments now get integration verification. \`\"poc\"\`, \`\"proof of concept\"\`, \`\"demo\"\`, and \`\"test\"\` still skip.

### New — \`TestGenerateContractsByDomain\` (4 tests)

- \`test_generates_contracts_keyed_by_domain\` — happy path, 2 domains × 5 LLM calls = 10, files on disk, keyed by domain name
- \`test_empty_domains_returns_empty_dict\` — early-return path, LLM not instantiated
- \`test_domain_with_empty_response_returns_none_for_that_domain\` — empty-response filter
- \`test_generate_design_content_still_produces_same_output_shape\` — end-to-end regression: task-keyed adapter still produces the same output as before the refactor

### Updated — \`TestShouldAddIntegrationTask\`

- \`test_experiment_skips_integration\` → \`test_experiment_gets_integration\` (assertion flipped to reflect new correct behavior)
- Added \`test_test_project_still_skips_integration\` to confirm other skip keywords still work

### Unchanged

- All 7 existing \`TestDesignAutocompleteParallelism\` tests pass with the refactored internals (this is the critical regression check — the task-keyed entry point's externally-observable behavior is unchanged)

## Quality gates

| Check | Result |
|---|---|
| \`pytest -m unit\` | **390 passed** (was 386; +4 from new tests), 0 failed |
| \`pytest tests/integration -m \"integration and internal\"\` | **7 passed**, 0 failed |
| \`mypy --strict\` on changed source files | clean |
| \`black\` + \`isort\` | applied |

## What this PR does NOT do

Explicitly out of scope — all of this is **PR 2 territory**:

- Does not move the call site of contract generation (still runs in the existing background \`_run_design_phase\` via \`_generate_design_content\`)
- Does not introduce the contract-first decomposer or feature flag
- Does not add the \`responsibility\` field to the Task model
- Does not modify agent prompt generation

## Diff stats

\`\`\`
src/integrations/integration_verification.py       |   7 +/-
src/integrations/nlp_tools.py                      | 341 +/-
tests/unit/integrations/test_design_autocomplete_parallelism.py | 187 +/-
tests/unit/integrations/test_integration_verification.py  |  29 +/-
4 files changed, 465 insertions(+), 99 deletions(-)
\`\`\`

Most of the diff in \`nlp_tools.py\` is docstrings — the actual logic delta is small.


lwgray:
## Review triage

Thanks for the thorough passes from both reviewers. Triaged below.

### ✅ Codex P2 — Preserve all design tasks when building domain map (FIXED in 7ca0ee2)

You're right and this is a real regression. Walked through the failure mode:

- **Pre-#320 behavior:** per-task iteration. If two tasks happen to share a stripped domain name (say, both ``"Design Authentication"`` with different descriptions), both tasks get their own LLM calls and both are marked DONE.
- **First version of this PR:** dict-keyed. Building ``domains[domain_name] = task.description`` silently drops the earlier write. Only one LLM call fires. Only one task is marked DONE. Two regressions.
- **Fixable-today?** Not reachable from the current PRD parser (``_create_bundled_design_tasks`` iterates a unique-keyed ``domains: Dict[str, List[str]]`` so collisions are impossible there), but nothing downstream enforces that invariant and the pre-#320 code was correct under the weaker assumption.

**Fix:** ``_generate_design_content`` no longer delegates to ``_generate_contracts_by_domain``. It iterates design tasks directly and calls ``_process_design_domain`` once per task, preserving pre-#320 behavior exactly. ``_generate_contracts_by_domain`` remains the public domain-keyed entry point for PR 2's contract-first decomposer — that path operates on a ``Dict[str, str]`` where uniqueness is a dict invariant, so the collision case is unreachable there.

**Regression test added:** ``test_duplicate_design_task_names_both_get_llm_calls_and_done``. Constructs two design tasks with the exact same name, asserts 10 LLM calls fire (not 5), asserts both are marked DONE, asserts both have the ``auto_completed`` label.

**Quality gates after the fix:**
- ``pytest -m unit`` → **391 passed** (was 390, +1 regression test), 0 failed
- ``mypy --strict`` clean
- Pushed as 7ca0ee2

### Triage of the manual review comments

**Minor concerns flagged, not fixed — with reasoning:**

1. **"Error Handling: _generate_single_decisions could benefit from more specific error types"** — noted but deferring. The current handler catches ``(json.JSONDecodeError, ValueError)`` which is the correct precise scope — we want to fall through to "empty decisions list" for any JSON parse failure but let other exceptions (like ``IntegrationError`` from retry exhaustion) propagate. More granular handling doesn't change behavior; it only changes log messages.

2. **"Resource Cleanup: _cleanup_background could accumulate warnings in production"** — out of scope for this PR (the function isn't touched by the refactor) and vague (no specific failure mode described). Leaving for a separate observability pass if warnings start accumulating in practice.

3. **"Metrics/logging for domain processing times"** — future enhancement, not a blocker. The #304 smoke-test script (\`scripts/smoke_test_design_autocomplete.py\`) already captures wall-clock at the function level. Per-domain latency would be useful for tuning the semaphore value but isn't required for PR 1.

4. **"Configuration option for artifact file organization"** — speculative. The ``ARTIFACT_PATHS`` dict in ``src/marcus_mcp/tools/attachment.py`` already exists and is mutable. Adding env-var overrides is YAGNI until someone hits a real need.

5. **"Add similar documentation to the module docstring"** — fair, but the module docstring covers the module as a whole. The PR-specific context lives in the function docstrings (``_generate_contracts_by_domain``, ``_process_design_domain``, ``_generate_design_content``), which is the right layer for this level of detail.

### Final state

- **Codex P2:** fixed in 7ca0ee2
- **Manual review suggestions:** triaged, none are blockers
- **Tests:** 391 unit passing (was 386 on develop + 5 from this PR — 4 for the new domain-keyed entry point + 1 for the duplicate-names regression)
- **Ready for re-review.**

Issue #297, opened 2026-04-02
feat: auto-complete design tasks with AI-generated artifacts during create_project


lwgray:
## Problem

When a worker agent completes a Design task then immediately gets an Implementation task, it over-executes — building everything it just designed, not just its assigned piece. This is a **context contamination** problem: the design knowledge is hot in the agent's context window and can't be prompt-engineered away.

**Evidence from dashboard-v16 experiment (Epictetus audit):**
- agent_unicorn_2 was assigned "Design Dashboard" then "Implement Time Widget"
- Built BOTH widgets, backend, all tests, full scaffold — 1130 lines in one commit
- agent_unicorn_1 got "Implement Weather Widget" but code already existed — could only write docs
- Contribution distribution verdict: **Single-Author Product**
- Multi-agency effective: **false**

**Root cause confirmed:** The task descriptions were well-scoped ("implement a Time Widget component..."). The agent over-executed because it had just authored the full system design covering both widgets. The same agent cannot do design and implementation.

## Proposed Solution

Auto-complete design tasks during `create_project` so no worker ever carries design authorship context. Follows the existing About task pattern (`_create_about_task` in `nlp_tools.py:644` — task created with `status=TaskStatus.DONE`, `assigned_to=None`).

### How It Works

1. `create_project` decomposes spec into tasks (existing — step 1)
2. **NEW:** For each design task, an AI call generates the requested artifacts (architecture docs, API contracts, data models, integration specs) — driven by the task description, not complexity tier
3. Design tasks created on kanban board with `status=DONE` + artifacts stored in `docs/` via standard `ARTIFACT_PATHS`
4. Workers see design tasks as completed dependencies, read artifacts via `get_task_context`
5. Cato hides auto-completed design tasks from the main board view (filter on `"auto_completed"` label)

### What It Does NOT Change

- Marcus stays a coordinator — the AI call to produce design artifacts is the same category as the AI call that already decomposes tasks
- Workers still see design tasks on the board (visible, just pre-completed)
- `get_task_context` works identically — artifacts stored in same paths a worker would use
- Works across all 3 interfaces: `/marcus`, MCP direct, Posidonius

### Marcus Changes

1. **New module:** `src/integrations/design_autocomplete.py`
   - `autocomplete_design_tasks()` — identifies design tasks, generates artifacts via AI, marks DONE
   - AI prompt sends the design task's full description (which already specifies what artifacts to create per `advanced_parser.py:912-927`)
[...]

PR #326, opened 2026-04-11
fix(#320): reconnect Phase A→B design artifact registration

### LINKED ISSUE #320: feat: contract-first task decomposition (generate contracts before decomposing)

lwgray:
## Summary

Phase B (design artifact registration via MCP tools) has been dead code for 5 days. Design artifacts generated by Phase A never reached \`state.task_artifacts\`, so implementation tasks walking dependencies via \`_collect_task_artifacts\` found empty dicts for every design task dependency. Contract-first decomposition was silently non-functional.

**This PR is a side-quest bug fix.** It does NOT close #320. #320 PR 2 proper (\`decompose_by_contract\` + \`Task.responsibility\` + \`MARCUS_DECOMPOSER\` feature flag) is still ahead as a separate PR.

## Root Cause

Timeline:

- **2026-04-02** (#297, commit \`4daccb7\`): Two-phase design autocomplete landed. Phase A synchronous inside \`create_project_from_description\`, stored output in \`result[\"design_content\"]\`. Phase B in \`src/marcus_mcp/tools/nlp.py\` read the key after \`state.refresh_project_state()\`. Handoff worked.
- **2026-04-06** (#314, commit \`1c5c7f7\`): Phase A moved to background closure via \`asyncio.ensure_future\` to prevent Claude Code timeouts on long LLM calls. The line \`result[\"design_content\"] = design_content\` was deleted as part of the refactor. Phase B in nlp.py continued to read \`result.get(\"design_content\", {})\` — always \`{}\`. Dead code.
- **2026-04-06 → 2026-04-11**: Marcus generated design artifacts to disk on every project creation but never registered them in \`state.task_artifacts\`. Agents saw no contracts from design deps.
- **2026-04-11** (this PR): Regression discovered while planning #320 PR 2.

No existing test caught the regression. The \`TestRegisterDesignViaMcp\` unit tests mock \`design_content\` directly, bypassing the handoff.

## Fix

Consolidate Phase A and Phase B into a single module-level \`_run_design_phase\` function that runs both in lockstep inside the background task. Cannot break again without touching both phases together.

### Ordering Invariant

**Phase B registration MUST run before the kanban DONE update.** The DONE update unblocks implementation tasks from hard dependencies — if Phase B runs after, there's a sub-second window where an impl task unblocks before \`state.task_artifacts\` is populated, and the agent's dependency walk returns empty.

Ordering is pinned in \`_run_design_phase\` and tested by \`test_phase_b_registration_runs_before_kanban_done_update\`.

### \`src/integrations/nlp_tools.py\` (+247 / -73)

- New module-level async \`_run_design_phase(state, kanban_client, safe_tasks, created_tasks, description, project_name, project_root)\` — orchestrates Phase A → Phase B → kanban DONE → scaffold in a fixed order with load-bearing ordering
- \`NaturalLanguageProjectCreator.__init__\` takes optional \`state\` param (None default for backward compat)
- Closure inside \`create_project_from_description\` replaced with direct call to \`_run_design_phase\` via \`ensure_future\`
- Fail-fast semantics: Phase A exception short-circuits the entire phase (no partial state.task_artifacts, no kanban updates, no scaffold)

### \`src/marcus_mcp/tools/nlp.py\` (-30 / +13)

- Dropped the dead Phase B block that read \`result[\"design_content\"]\`
- Both call sites now pass \`state=state\` to \`NaturalLanguageProjectCreator\`

### \`tests/unit/integrations/test_design_autocomplete.py\` (+340)

New \`TestRunDesignPhaseHandoff\` class with 4 chain-level regression tests:

1. \`test_run_design_phase_populates_state_task_artifacts\` — full chain: Phase A → Phase B → \`state.task_artifacts[design_task_id]\` populated. This is the assertion that would have caught #314's regression on day one.
2. \`test_phase_b_registration_runs_before_kanban_done_update\` — ordering invariant: Phase B \`log_artifact\` fires before \`kanban.update_task\` DONE
3. \`test_run_design_phase_handles_phase_a_failure\` — Phase A exception short-circuits, no Phase B, no kanban updates, no partial state
4. \`test_run_design_phase_noop_when_no_design_tasks\` — empty \`design_content\` is a clean no-op

### \`docs/source/systems/coordination/52-design-autocomplete.md\` (rewritten)

The old version still described the pre-#314 design (Phase A synchronous, tasks born DONE). Updated to match current code:

- Background execution model via \`ensure_future\`
- \`assigned_to=\"Marcus\"\` race mitigation (replaces \"born DONE\" invariant)
- Ordering invariant (Phase B before kanban DONE)
- **Regression History** section documenting the 5-day outage so future refactors know why the two phases were consolidated

## Test Results

- \`tests/unit/integrations/test_design_autocomplete.py\`: 15/15 passed
- \`pytest -m unit\`: 391/391 passed
- \`mypy src/integrations/nlp_tools.py src/marcus_mcp/tools/nlp.py\`: clean

## Test plan

- [x] Chain-level regression test: create_project → background Phase A → Phase B → \`state.task_artifacts\` populated
- [x] Ordering invariant test: Phase B log_artifact before kanban update_task DONE
- [x] Phase A failure: no partial state, no kanban updates, no scaffold
- [x] Empty design_content: clean no-op
- [x] Backward compat: optional \`state\` param defaults to None, legacy behavior preserved
- [x] \`pytest -m unit\` green
- [x] mypy clean on touched files
- [ ] Post-merge: run snake game with Phase B alive, verify integration_verification.py receives richer evidence

## Related

- Supersedes nothing (new PR)
- Related to #320 PR 2 (does NOT close — PR 2 proper is decomposer work still ahead)
- Regression from #314
- Built on #297's infrastructure

### DIFF
--- a/docs/source/systems/coordination/52-design-autocomplete.md
+++ b/docs/source/systems/coordination/52-design-autocomplete.md
@@ -5,8 +5,8 @@
 | Field | Value |
 |-------|-------|
 | Status | Implemented |
-| Version | 1.0 |
-| Date | 2026-03-31 |
+| Version | 2.0 |
+| Date | 2026-04-11 |
 
 ## Problem
 
@@ -18,29 +18,53 @@ The root cause is that the agent has the full system design in its context when 
 
 ## Solution
 
-Marcus auto-completes design tasks during `create_project`, before any agent is spawned. Design artifacts are produced by Marcus itself using targeted LLM calls, and design tasks arrive on the board already marked DONE. No agent ever touches a design task.
+Marcus auto-completes design tasks during `create_project` without letting any agent touch them. Design artifacts are produced by Marcus itself using targeted LLM calls. The implementation runs in two phases that together live inside a single background task, `_run_design_phase`:
 
-This is a two-phase process because it spans two different stages of project creation.
+- **Phase A** generates design artifacts to disk via parallel LLM calls.
+- **Phase B** registers those artifacts into the MCP `state.task_artifacts` so downstream implementation tasks discover them via `get_task_context`.
 
-## Phase A: Generate Content (Before Board Creation)
+Both phases must run in lockstep. They used to be wired independently and that split broke silently when one was refactored — see [Regression History](#regression-history) below.
 
-Runs inside `create_project`, before `create_tasks_on_board()` is called.
+## Race Mitigation: `assigned_to="Marcus"`
+
+Design tasks are born with `assigned_to="Marcus"` rather than born DONE. Marcus's task assignment filter treats any task whose `assigned_to` is `"Marcus"` as off-limits to agents regardless of status. This prevents agents from grabbing design tasks during the window between `create_tasks_on_board()` and the kanban DONE update that marks Phase A complete.
+
+This is the mechanism that replaced the original "born DONE" invariant when Phase A moved to a background task in GH-314. The race still exists at the board level (tasks are TODO for a few seconds), but the assignment filter closes it at the allocation layer.
+
+## Background Execution Model
+
+`_run_design_phase` runs as an `asyncio.ensure_future` background task, scheduled after `create_tasks_on_board()` returns and BEFORE `create_project_from_description()` returns to its caller. The MCP tool response is not blocked on Phase A's LLM calls (1–3 min wall-clock after the GH-304 parallelization). Implementation tasks remain blocked on hard dependencies until the kanban DONE update inside the closure fires.
+
+```
+create_project_from_description():
+  ├── parse_prd_to_tasks() [synchronous]
+  ├── create_tasks_on_board() [synchronous]
+  │     design tasks born with assigned_to="Marcus", status=TODO
+  ├── asyncio.ensure_future(_run_design_phase(...))  [fire-and-forget]
+  └── return result  [MCP tool responds immediately]
+
+_run_design_phase() [background]:
+  ├── Phase A: _generate_design_content()
+  │     └── parallel LLM calls, write files to disk
+  ├── Phase B: _register_design_via_mcp()         ← MUST be before kanban DONE
+  │     └── log_artifact() populates state.task_artifacts[design_task_id]
+  ├── Kanban DONE update                           ← unblocks impl tasks
+  │     └── kanban_client.update_task(id, "done")
+  └── _generate_project_scaffold()
+        └── writes initial project scaffolding
+```
+
+## Phase A: Generate Content
 
 For each design task, Marcus makes **separate LLM calls** for each artifact document:
 
 1. Architecture document
 2. API contracts
 3. Data models
-4. One call for decisions
+4. Interface contracts
+5. One call for decisions
 
 Each call writes its artifact file to disk at the standard `ARTIFACT_PATHS` location.
-
-Then Marcus sets on the task object:
-- `status = DONE`
-- `assigned_to = "Marcus"`
-- Label: `"auto_completed"`
-
-When `create_tasks_on_board()` runs, design tasks hit the board already DONE. No agent can grab them.
 
 ### Why separate LLM calls per artifact
 
@@ -48,51 +72,52 @@ A single LLM call producing all artifacts as one JSON response truncated at 7,71
 
 Separate calls per document mirror how agents actually work. Each document is a focused, bounded response that fits comfortably within output limits.
 
-## Phase B: Register Metadata (After State Refresh)
+### Parallelism (GH-304)
 
-Runs inside `src.marcus_mcp.tools.nlp.create_project()`, after `state.refresh_project_state()` has populated the MCP state object.
+The five LLM calls per task run concurrently (Level 2), and all design tasks run concurrently with each other (Level 1). A per-invocation `asyncio.Semaphore(10)` caps total in-flight LLM calls to respect provider rate limits. Wall-clock: 25–33 min sequential → 1–3 min parallel on a typical 10-task enterprise project.
 
-Phase B registers the artifacts and decisions through MCP tools:
+### Fail-fast semantics
 
-- `log_artifact()` for each file produced in Phase A — populates `state.task_artifacts`
+Each LLM call is wrapped with `@with_retry(max_attempts=3, base_delay=2.0, jitter=True)`. If retries exhaust on any single call, the exception propagates through the inner `gather`, aborts the outer `gather`, and `_generate_design_content` raises. `_run_design_phase` catches this, logs it, and short-circuits: no Phase B, no kanban updates, no scaffold, no partial state. Fail-fast was chosen over warn-and-continue because partial design outputs silently corrupt downstream agent work.
+
+## Phase B: Register Metadata
+
+Phase B registers artifacts and decisions through MCP tools:
+
+- `log_artifact()` for each file produced in Phase A — populates `state.task_artifacts[design_task_id]`
 - `log_decision()` for each architectural decision — populates `state.context.decisions`
 
-Phase B does **NOT** call `report_task_progress`. The task is already DONE from Phase A.
+Phase B does **NOT** call `report_task_progress`. The task's kanban DONE update happens in the next step.
 
-Workers discover design artifacts and decisions through `get_task_context` when they start dependent tasks.
+Workers discover design artifacts and decisions through `get_task_context` when they start dependent tasks. The retrieval path is `_collect_task_artifacts` (context.py) which walks `task.dependencies`, pulls `state.task_artifacts[dep_id]` for each, and returns the merged list to the agent.
 
-## Why Two Phases
+## Ordering Invariant
 
-```
-create_project flow:
+**Phase B registration MUST run before the kanban DONE update.** This is a load-bearing ordering constraint pinned by `_run_design_phase` and tested by `TestRunDesignPhaseHandoff`.
 
-  ┌──────────────────────────┐
-  │  Phase A                 │
-  │  - LLM calls             │   Needs to run BEFORE create_tasks_on_board()
-  │  - Write files to disk   │   so tasks are born DONE (prevents race
-  │  - Set status = DONE     │   condition where agents grab design tasks)
-  │  - Set auto_completed    │
-  └──────────┬───────────────┘
-             │
-             v
-  ┌──────────────────────────┐
-  │  create_tasks_on_board() │   Design tasks arrive DONE
-  └──────────┬───────────────┘
-             │
-             v
-  ┌──────────────────────────┐
-  │  refresh_project_state() │   Populates MCP state object
-  └──────────┬───────────────┘
-             │
-             v
-  ┌──────────────────────────┐
-  │  Phase B                 │   Needs MCP state object, which is only
-  │  - log_artifact()        │   available after refresh_project_state()
-  │  - log_decision()        │
-  └──────────────────────────┘
-```
+The kanban DONE update is what unblocks implementation tasks from hard dependencies. If Phase B runs after, there is a window where:
 
-The two phases serve different purposes and cannot be combined. Phase A needs to complete before board creation to prevent the race condition. Phase B needs the MCP `state` object which only exists after `refresh_project_state()`.
+1. Design task marked DONE on kanban
+2. Implementation task unblocked
+3. Agent requests implementation task
+4. `_collect_task_artifacts` walks dependencies, finds empty `state.task_artifacts[design_task_id]`
+5. Agent receives no contracts, proceeds without them
+
+The window is sub-second but races don't care about narrow windows. Phase B before kanban update closes it entirely.
+
+## Regression History
+
+This section exists because the Phase A → Phase B handoff was broken for 5 days (2026-04-06 to 2026-04-11) and the regression hid behind stale documentation.
+
+**2026-04-02 (GH-297, commit 4daccb7):** Two-phase design autocomplete landed. Phase A ran synchronously inside `create_project_from_description` and stored output in `result["design_content"]`. Phase B ran in `src/marcus_mcp/tools/nlp.py` after `state.refresh_project_state()`, reading the key from the result dict.
+
+**2026-04-06 (GH-314, commit 1c5c7f7):** Phase A moved to a background closure via `asyncio.ensure_future` to prevent Claude Code from timing out on long LLM calls. The line `result["design_content"] = design_content` was deleted as part of the refactor. Phase B in nlp.py continued to read `result.get("design_content", {})` but the key was never populated again. Phase B became dead code. No test caught it — the `TestRegisterDesignViaMcp` unit tests mocked `design_content` directly, bypassing the handoff.
+
+**2026-04-06 to 2026-04-11:** Marcus generated design artifacts to disk on every project creation but never registered them in `state.task_artifacts`. Agents calling `get_task_context` on implementation tasks walked the dependency graph and found empty artifact lists for every design task dependency. Contract-first decomposition was silently non-functional.
+
+**2026-04-11 (GH-320):** Regression discovered while planning the contract-first decomposer. Phase A and Phase B consolidated into a single `_run_design_phase` function that cannot break again without touching both phases. The dead Phase B block in nlp.py was removed. The ordering invariant (Phase B before kanban DONE) was pinned by `TestRunDesignPhaseHandoff`. This document was rewritten to match current code.
+
+**Lesson:** Cross-cutting handoffs between code owned by the same closure are fragile when the closure is refactored. If you ever split Phase A and Phase B again — or move Phase A to a different lifecycle hook — preserve the ordering invariant and add a chain-level test that exercises the full handoff with both phases mocked.
 
 ## Why MCP Tools for Registration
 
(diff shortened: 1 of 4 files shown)


lwgray:
## Review response

Addressed three of the priority items in commit c18bc02. Left three items out of scope with explanation.

### Fixed

**1. O(n²) kanban update loop** — replaced `created_tasks.index(ct)` inside the loop with `zip(created_tasks, safe_tasks)`. The two arrays are index-aligned by construction, so this is both correct and O(n). Bonus: `zip` handles desynchronized arrays gracefully by stopping at the shorter list.

**2. Type annotations** — `safe_tasks` and `created_tasks` are now `List[Task]` instead of `List[Any]`. The `Task` import was already available in the module.

**3. Test for state=None legacy backward-compat path** — the PR docstring claimed optional `state` preserves legacy behavior but nothing tested it. New test `test_run_design_phase_with_state_none_skips_phase_b` verifies: Phase B is skipped cleanly, Phase A kanban DONE updates still fire, scaffold still runs. That's the full legacy contract.

**4. Bonus test for mismatched task array lengths** — defense-in-depth regression guard. New test `test_run_design_phase_handles_mismatched_task_arrays` pins the `zip` behavior: if a future refactor ever desynchronizes the arrays, `zip` stops at the shorter list rather than raising IndexError. Test uses a pathologically short `created_tasks` and asserts `update_task` is called exactly once.

### Not addressed (out of scope for this PR)

**Granular exception handling** — this is a background task where the failure mode that matters is "did we leave partial state?" Coarse `except Exception` with fail-fast semantics is correct for that mode. Retry logic belongs at the resilience layer, not scattered across the background task.

**Concurrent `_run_design_phase` calls** — there's exactly one background task per `create_project` call. No concurrent invocations. Not a real risk.

**Retry logic for transient kanban API failures** — separate concern. The kanban client itself should handle retries. Adding retry logic here couples the design phase to kanban transport concerns.

### Test results

- 17/17 design_autocomplete tests pass (15 original + 2 new)
- 391/391 unit tests pass
- mypy clean on touched files

Ready for re-review or merge.

lwgray:
## Codex P1 addressed (commit 2df64ce)

Good catch on the zero-registration race. You're right — the ordering fix alone didn't close it, because Phase B could succeed with `artifacts_registered=0` if `state.project_tasks` was stale when the background task fired. That reintroduces the exact silent-failure mode this PR was built to eliminate.

### Fix

Two complementary guards:

**1. Pre-Phase-B state refresh** — `_run_design_phase` now calls `state.refresh_project_state()` before Phase B (if the method exists on state). This closes the race window between `ensure_future` and the MCP tool caller's own refresh. Phase B gets a current view of kanban UUIDs rather than whatever was on state when the background task was scheduled.

**2. Zero-registration guard** — after Phase B returns, the closure checks `artifacts_registered`. If Phase A produced `design_content` but Phase B registered zero artifacts, we log an ERROR and skip the kanban DONE update entirely:

```python
if state is not None and design_content and phase_b_registered == 0:
    logger.error(
        "[design_autocomplete] Phase B registered 0 artifacts "
        "despite Phase A producing design_content. Refusing to "
        "mark design tasks DONE — impl tasks would unblock "
        "without contracts (exact silent failure mode from #314). "
        ...
    )
    return
```

Design tasks stay TODO — the correct degenerate state. Loud failure > silent corruption. The user sees "design tasks never completed" on the kanban board and can investigate, rather than "implementation agents produced code that doesn't integrate and nobody knows why."

### Why not your exact fix

Your suggestion was to return early on zero registrations. I did that, but added the pre-refresh step too because the root cause of zero registrations is usually the stale-state race. Catching the symptom (zero registrations → skip DONE) is a backstop; refreshing state before Phase B is the actual fix that should make the backstop rarely fire.

### Exemption for `state is None`

The legacy backward-compat path (`state is None`) is exempt from the zero-registration guard. Those callers explicitly opt out of Phase B entirely — for them, the DONE updates are the only thing moving design tasks to the next kanban state. Without the exemption, legacy callers would see their projects hang with design tasks stuck in TODO.

### New tests

- **`test_run_design_phase_skips_done_when_phase_b_registered_zero`** — simulates the race by giving Phase B empty `state.project_tasks`. Asserts kanban `update_task` is NOT called and scaffold is NOT run. This is the direct regression guard for your finding.
- **`test_run_design_phase_refreshes_state_before_phase_b`** — asserts `state.refresh_project_state` is called once before Phase B runs.

One existing test (`test_run_design_phase_handles_mismatched_task_arrays`) needed to populate `state.project_tasks` so Phase B could match at least one task by name — otherwise the new zero-registration guard fires and short-circuits. That's not a test bug; it's the guard working correctly.

### Test results

- 19/19 design_autocomplete tests pass (17 previous + 2 new)
- mypy clean

Ready for re-review.

PR #327, opened 2026-04-11
feat(#320): contract-aware decomposer + Task.responsibility (PR 2)

### LINKED ISSUE #320: feat: contract-first task decomposition (generate contracts before decomposing)

lwgray:
## Summary

Contract-first task decomposition per GH-320 acceptance criteria. Experiment 1 (2026-04-10, hand-crafted TypeScript contract, snake game) validated the mechanism: ~30/70 contribution split, clean merge, first Marcus experiment on tight coupling that did not produce a Single-Author Product verdict. **This PR productionizes the path.**

Does NOT close #320 — Experiment 4 (LLM-generated contract snake game) is the real validation gate. Will be run after this PR merges.

### 1. \`Task.responsibility\` field (src/core/models.py)

Optional Task attribute naming the contract interface this task owns, e.g. \`"implements GameEngine interface from src/types.ts"\`. Orthogonal to \`provides\`/\`requires\`:

- \`provides\`/\`requires\` describe what the task satisfies at the dependency-wiring layer
- \`responsibility\` names the contract interface the task owns from the authoritative shared contract file

### 2. \`AdvancedPRDParser.decompose_by_contract\` (src/ai/advanced/prd/advanced_parser.py)

New method signature:

\`\`\`python
async def decompose_by_contract(
    self,
    prd_analysis: PRDAnalysis,
    contract_artifacts: Dict[str, Optional[Dict[str, Any]]],
    agent_count: int,
    constraints: Optional[ProjectConstraints] = None,
) -> List[Task]
\`\`\`

Takes PRD analysis + contract artifacts (from PR 1's \`_generate_contracts_by_domain\`) + agent count. Structured LLM output via \`AIAnalysisEngine.generate_structured_response\` produces task list where each task carries \`responsibility\`, \`contract_file\` in \`source_context\`, \`provides\`, \`requires\`, and \`estimated_minutes\`. Raises \`ValueError\` on all-empty contracts and \`RuntimeError\` on LLM failure so the caller can fall back gracefully.

### 3. \`MARCUS_DECOMPOSER\` feature flag (src/config/decomposer_config.py)

New module with \`resolve_decomposer(options)\` and \`is_contract_first(options)\`. Precedence:
1. \`options["decomposer"]\` (explicit per-call override)
2. \`MARCUS_DECOMPOSER\` env var
3. Default: \`feature_based\`

Unknown strategy values trigger a loud warning and fall back to \`feature_based\`. No silent fallbacks.

### 4. \`NaturalLanguageProjectCreator.process_natural_language\` (src/integrations/nlp_tools.py)

Branches on the flag. Contract-first path runs:
1. Deep PRD analysis (\`_analyze_prd_deeply\`)
2. Domain discovery (\`_discover_domains\`)
3. Contract generation (\`_generate_contracts_by_domain\`)
4. Contract-aware decomposition (\`decompose_by_contract\`)

Any failure at any stage returns \`None\` from the helper, and the caller falls back to feature-based with a visible warning. Fallback triggers: \`project_root\` missing, PRD analysis fails, no functional requirements, domain discovery fails, no domains discovered, contract generation fails, all contracts empty/None, decomposer raises \`ValueError\`/\`RuntimeError\`, decomposer returns empty list.

### 5. Agent prompt Layer 1.3 (src/marcus_mcp/tools/task.py)

New \`CONTRACT RESPONSIBILITY\` layer in \`build_tiered_instructions\`. Fires when \`Task.responsibility\` is set:

\`\`\`
📜 CONTRACT RESPONSIBILITY (contract-first decomposition):
You OWN: implements GameEngine interface from types.ts
Contract file: docs/api/types.ts

BEFORE writing any code, Read() the contract file at docs/api/types.ts.
...
Your implementation MUST conform to the contract. If the contract is
missing something you need, report a blocker — do NOT silently modify
the contract file.
\`\`\`

Fires BEFORE the subtask context layer because contract ownership is structural — an agent needs to understand which contract interface it owns before considering whether this is a subtask of a larger parent.

Also: \`AIAnalysisEngine.generate_task_instructions\` adds \`responsibility\` + \`contract_file\` + \`contract_first: true\` to the \`task_data\` dict so the LLM sees them when generating base instructions.

### 6. Integration task contract authority (src/integrations/integration_verification.py)

\`IntegrationTaskGenerator.create_integration_task\` and \`enhance_project_with_integration\` take an optional \`contract_file\` parameter. When set, the task description prepends a \`CONTRACT-FIRST PROJECT\` preamble:

> The contract is AUTHORITATIVE. Agents built their implementations against it. If any implementation diverges from the contract during your verification, FIX THE IMPLEMENTATION — do NOT modify the contract file.

Prevents the integration agent from silently \"fixing\" a mismatch by editing the contract, breaking the invariant that made contract-first decomposition work in the first place.

\`nlp_tools.process_natural_language\` picks up the contract file from the first contract-first task's \`source_context\` and passes it through \`enhance_project_with_integration\`.

### \`tests/unit/config/test_decomposer_config.py\` (11 tests)

Precedence order, unknown strategy fallback with warning, \`is_contract_first\` convenience wrapper.

### \`tests/unit/ai/test_decompose_by_contract.py\` (8 tests)

- \`test_happy_path_produces_tasks_with_responsibility\` — structured LLM → Task objects with responsibility set
- \`test_empty_contracts_raises_value_error\` — all-None → ValueError
- \`test_contracts_with_empty_artifacts_raises_value_error\` — \`{\"artifacts\": []}\` → ValueError
- \`test_llm_failure_raises_runtime_error\` — LLM exception → RuntimeError
- \`test_empty_task_list_in_response_raises\` — \`{\"tasks\": []}\` → RuntimeError
- \`test_labels_include_contract_first\` — tasks carry \`contract_first\` label
- \`test_contract_file_embedded_in_description\` — agents can discover contract from task.description
- \`test_mixed_none_and_usable_contracts\` — partial contracts → decomposer uses only usable ones

### \`tests/unit/integrations/test_contract_first_fallback.py\` (9 tests)

All failure paths for \`_try_contract_first_decomposition\` return \`None\` so the caller falls back to feature_based. Plus one happy path.

### \`tests/unit/mcp/test_contract_responsibility_prompt.py\` (6 tests)

- Contract responsibility layer fires when set
- Contract file named in instructions
- No layer when responsibility unset (legacy tasks)
- Layer forbids silent contract modification
- Layer fires BEFORE subtask layer (ordering)
- Graceful fallback when contract_file missing

### \`tests/unit/integrations/test_integration_verification.py\` (5 new)

- \`test_integration_task_includes_contract_preamble_when_set\`
- \`test_integration_task_preamble_forbids_contract_modification\`
- \`test_integration_task_no_preamble_when_contract_file_none\` (backward compat)
- \`test_enhance_project_with_integration_passes_contract_file\`
- \`test_enhance_project_with_integration_default_no_contract\`

## Test results

- \`pytest -m unit\`: **425/425 passed** (391 baseline + 34 new)
- mypy clean on all touched files

## Scope explicitly excluded (future work)

- **Experiment 2** (Python log analyzer): pending per #320
- **Experiment 3** (3-agent contract-based decomposition): pending — unknown territory for boundary discipline
- **Experiment 4** (LLM-generated contract snake game): pending — the real productionization validation gate, this is what proves LLM-generated contracts (not hand-crafted) can drive automatic decomposition
- **Contract amendment flow**: what happens when an agent discovers mid-implementation that the contract is missing something they need? Not scoped in this PR — follow-up issue

## Related

- GH-320 acceptance criteria: PR 2 items landed (decompose_by_contract, Task.responsibility, agent prompt, integration task, feature flag, fallback)
- GH-322: PR 1 (domain-keyed contract generation) — prerequisite
- GH-326: reconnect Phase A→B handoff — prerequisite (merged)
- Experiment 1 (2026-04-10): hand-crafted contract validation on snake game

## Test plan

- [x] Unit tests for decomposer (8)
- [x] Unit tests for feature flag precedence (11)
- [x] Unit tests for fallback on every failure path (9)
- [x] Unit tests for agent prompt surfacing (6)
- [x] Unit tests for integration task contract authority (5)
- [x] \`pytest -m unit\` green (425/425)
- [x] mypy clean on touched files
- [ ] **Experiment 3** (3-agent contract-based): post-merge
- [ ] **Experiment 4** (LLM-generated contract snake game): post-merge, this is the real validation gate

### DIFF
--- a/docs/experiments/gh320-experiment-3-runbook.md
+++ b/docs/experiments/gh320-experiment-3-runbook.md
@@ -1,0 +1,145 @@
+# GH-320 Experiment 3 — 3-agent contract-based decomposition
+
+| Field | Value |
+|-------|-------|
+| Status | Not yet run |
+| Blocked by | Nothing (can run any time after PR #327 merges) |
+| Motivation | Test whether contract-based decomposition maintains boundary discipline and clean merges when three agents work in parallel on a tightly-coupled problem, not just two |
+| Success criterion | Contribution distribution avoids Single-Author Product verdict; clean merge across 3 worktrees; zero contract modifications |
+| Related | GH-320 PR 2, Experiment 1 (2026-04-10, snake game, 2 agents, hand-crafted contract) |
+
+## Why this experiment exists
+
+Experiment 1 validated 2-agent contract-based decomposition on a tightly-coupled problem. Two agents produced a ~30/70 contribution split, clean merge, zero contract modifications — the first Marcus experiment on tight coupling that did not produce a Single-Author Product verdict.
+
+**Unknowns at 3+ agents:**
+- Does boundary discipline hold when the contract has three interface boundaries instead of two?
+- Do transitive dependencies between contract interfaces cause cascading coordination failures?
+- Does the "file ownership emerges from contract ownership" invariant still hold when three agents touch three distinct files?
+- Does the contribution distribution stay balanced (33/33/34) or collapse into a Single-Author pattern?
+
+Experiment 1 alone is not sufficient evidence to generalize contract-first decomposition to N-agent runs. This experiment is the second data point.
+
+## Setup
+
+Same Vite + React + TypeScript scaffold as Experiment 1. Three-way split along three natural contract boundaries:
+
+- **Agent A — GameEngine**: implements `GameEngine` class (core game loop, state transitions, collision detection)
+- **Agent B — Renderer**: implements `Renderer` class (pure rendering of GameState to DOM, no input handling)
+- **Agent C — Controller**: implements `Controller` class (input handling, binds keyboard to engine, manages timing/game loop)
+
+Contract file `src/types.ts` (hand-crafted, committed on main before any agent starts):
+
+```typescript
+export type Direction = 'up' | 'down' | 'left' | 'right';
+export type Coord = { x: number; y: number };
+
+export interface GameState {
+  snake: Coord[];
+  food: Coord;
+  score: number;
+  gameOver: boolean;
+}
+
+export interface GameEngine {
+  tick(): GameState;
+  setDirection(dir: Direction): void;
+  reset(): void;
+  getState(): GameState;
+}
+
+export interface Renderer {
+  /** Render the game state to a container element. Pure: no mutation of state. */
+  render(state: GameState, container: HTMLElement): void;
+  /** Show game-over overlay. */
+  showGameOver(score: number, container: HTMLElement): void;
+}
+
+export interface Controller {
+  /** Bind keyboard events and start the game loop. Returns a cleanup function. */
+  start(engine: GameEngine, renderer: Renderer, container: HTMLElement): () => void;
+}
+```
+
+Note: three interfaces with explicit dependency ordering. `Controller` composes both `GameEngine` and `Renderer`, so it has the most coordination surface. This is the failure mode to watch — does Agent C stay within its contract or drift into rewriting the engine or renderer?
+
+## Execution
+
+```bash
+# Prereqs: git, node 20+, Claude Code CLI
+mkdir -p ~/experiments/snake-contract-3agent
+cd ~/experiments/snake-contract-3agent
+
+# Baseline
+npm create vite@latest baseline -- --template react-ts
+cd baseline
+# Copy types.ts above to src/types.ts
+git init && git add -A && git commit -m "baseline + contract"
+cd ..
+
+# Three worktrees off main
+git -C baseline worktree add ../agent-a-engine main
+git -C baseline worktree add ../agent-b-renderer main
+git -C baseline worktree add ../agent-c-controller main
+```
+
+Run three Claude Code sessions in parallel:
+
+**Agent A prompt:**
+> Implement the `GameEngine` interface from `src/types.ts` in a new file `src/gameEngine.ts`. Do not touch any other file. Do not modify `src/types.ts`. The contract is read-only and authoritative.
+
+**Agent B prompt:**
+> Implement the `Renderer` interface from `src/types.ts` in a new file `src/renderer.ts`. Render the snake game state to a container element using simple DOM manipulation (divs with inline styles) or canvas. Do not touch any other file. Do not modify `src/types.ts`.
+
+**Agent C prompt:**
+> Implement the `Controller` interface from `src/types.ts` in a new file `src/controller.ts`. Bind arrow key events on the container, call `engine.setDirection()` on input, drive the game loop with `setInterval`, and call `renderer.render()` on each tick. Also wire up `src/App.tsx` and `src/main.tsx` to instantiate `GameEngine`, `Renderer`, and `Controller` and mount them to a container. Do not touch `gameEngine.ts`, `renderer.ts`, or `types.ts`.
+
+## Measurements to capture
+
+After all three agents complete, merge the worktrees and verify:
+
+1. **Contribution distribution**: use Epictetus audit to check for Single-Author Product verdict. Target: ~33/33/34 split by lines of product code. Hard failure if any agent produced <10% of total.
+2. **Contract adherence**: `git diff main..HEAD -- src/types.ts` should be empty. Any diff = contract modification = failed experiment.
+3. **File boundary discipline**: each agent's worktree should only modify its owned file plus the shared scaffolding files in Agent C's case (App.tsx, main.tsx). Cross-boundary edits = boundary discipline failure.
+4. **Clean merge**: three-way merge via octopus merge or sequential merges should produce zero conflicts. Contract-based boundaries should prevent conflicts by construction.
+5. **Build status**: `tsc --noEmit` exit 0 after merge.
+6. **Runtime check**: `npm run dev`, hit arrow keys, verify snake moves, food spawns, game-over fires.
+
+## Expected outcomes and interpretation
+
+### Hypothesis A (contract discipline holds at N=3)
+
+Contribution distribution is roughly balanced. All three agents produce substantive code in their owned files. No contract modifications. Clean three-way merge. Snake game runs.
+
+**Implication**: contract-based decomposition generalizes beyond 2 agents. Marcus can safely recommend contract-first for projects with 3+ agents on tightly-coupled problems.
+
+### Hypothesis B (Single-Author Product re-emerges)
+
+One agent (likely Agent C, the one with the most coordination surface) absorbs all the work. Agent A and/or B produce empty or trivial files. Or one agent silently edits another's contract.
+
+**Implication**: contract-based decomposition has an N-agent scaling ceiling. Need to investigate whether the failure is:
+- Contract design (3 interfaces not enough isolation; need 4+ or different shape)
+- Prompt design (agents don't understand ownership)
+- Fundamental (contracts can't carry enough information to isolate 3+ agents)
+
+### Hypothesis C (partial success)
+
+Two agents produce substantive code, one produces empty or trivial. Contract is not modified. Merge is clean.
+
+**Implication**: closer to #267's original Single-Author observation than to experiment 1's result. Contract-based decomposition helps but doesn't fully solve N-agent coordination at N=3. Need to understand which contract interface failed to carry enough signal.
+
+## What to do with the result
+
+Whichever hypothesis fires, write the outcome to `docs/audit-reports/snake-contract-3agent-YYYY-MM-DD.json` in the Epictetus format used by prior audits. Update GH-320 with the finding. If hypothesis B or C fires, file a follow-up issue on what's next — likely "experiment 3b with 4 contract interfaces" or "contract amendment flow" (the open question from the #320 issue body).
+
+## What this experiment does NOT test
+
+- **LLM-generated contracts** — this uses a hand-crafted contract. Experiment 4 covers LLM-generated contracts.
+- **The Marcus contract-first decomposer** — this runs Claude Code agents directly, not via Marcus. Experiment 4 covers the full Marcus pipeline.
+- **Contract amendment mid-implementation** — if an agent discovers the contract is missing something, what happens? Out of scope. Separate experiment.
+
+## Related
+
+- GH-320 issue body, "Experiment 3" section
+- `~/experiments/snake-contract-test/RESULT.md` — Experiment 1 result template
+- Experiment 4 (`gh320-experiment-4-runbook.md`) — LLM-generated contract validation via Marcus pipeline
--- a/docs/experiments/gh320-experiment-4-runbook.md
+++ b/docs/experiments/gh320-experiment-4-runbook.md
@@ -1,0 +1,144 @@
+# GH-320 Experiment 4 — LLM-generated contract snake game (validation gate)
+
+| Field | Value |
+|-------|-------|
+| Status | Not yet run |
+| Blocked by | PR #327 must merge (contract-aware decomposer implementation) |
+| Motivation | Validate that contract-first decomposition works end-to-end through Marcus's actual pipeline with an LLM-generated contract, not a hand-crafted one |
+| Success criterion | Reproduce Experiment 1's ~30/70 contribution split and clean merge on snake game **without** a human writing the contract |
+| Related | GH-320, PR #327, Experiment 1 (2026-04-10, hand-crafted contract) |
+
+## Why this experiment exists
+
+This is **the real productionization validation gate for PR 2b**.
+
+Experiment 1 proved the mechanism: if you give two agents a good contract and task descriptions that tie each agent to one side of the contract, they produce integrated code. But the contract was **hand-crafted by a human**. Production Marcus will use **LLM-generated contracts** via `_generate_contracts_by_domain`.
+
+The weak link in contract-first decomposition is the contract itself. An LLM-generated contract can be:
+- Ambiguous (interface methods without semantics)
+- Incomplete (missing a method one agent needs)
+- Misaligned (types that don't actually compose)
+
+If the LLM-generated contract is bad, downstream decomposition produces tasks with vague responsibilities, agents produce code that doesn't integrate, and we're back to the Single-Author Product verdict. **Until Experiment 4 passes, PR #327 is unproven in production.**
+
+GH-320 issue body, verbatim: *"LLM-generated contracts are the weak link. Everything above assumes contract generation produces output good enough to decompose from. The experiment-1 contract was hand-crafted by a human. Production Marcus will use LLM-generated contracts. Failure mode: the contract is ambiguous, has type mismatches, or omits a method one agent needs."*
+
+## Prerequisites
+
+1. **PR #327 merged to develop**. Contract-aware decomposer, Task.responsibility field, feature flag, and agent prompt surfacing must all be live.
+2. **PR #326 merged to develop**. Phase A→B handoff must work so design artifacts flow to `state.task_artifacts` and downstream agents can discover them via `get_task_context`. (Merged 2026-04-11.)
+3. **Marcus configured with Anthropic API key**. Contract generation needs real LLM calls.
+4. **Clean baseline directory** for the experiment (no prior project state).
+
+## Setup
+
+```bash
+# Ensure develop is up to date
+cd /Users/lwgray/dev/marcus
+git checkout develop
+git pull origin develop
+
+# Verify both prereqs landed
+git log --oneline | head -5
+# Expect: feat(#320): contract-aware decomposer + Task.responsibility (PR 2)
+#         fix(#320): reconnect Phase A→B design artifact registration (#326)
+
+# Start Marcus with contract-first flag on
+export MARCUS_DECOMPOSER=contract_first
+# Start Marcus per your normal startup procedure
+```
+
+## Execution
+
+Create a new snake game project via Marcus with contract-first enabled:
+
+```python
+# Via Claude Code or direct MCP call:
+create_project(
+    description=(
+        "Build a classic snake game in TypeScript with React. "
+        "The snake moves on a grid, eats food, grows, and ends "
+        "the game on wall or self collision. Arrow keys control "
+        "direction. Display current score and a restart button "
+        "on game over."
+    ),
+    project_name="snake-game-exp4",
+    options={
+        "decomposer": "contract_first",  # Explicit override, belt-and-suspenders
+        "agent_count": 2,
+        "project_root": "/Users/lwgray/dev/snake-game-exp4",
+        "complexity": "standard",
+    }
+)
+```
+
+Then spawn 2 agents as usual. Observe:
+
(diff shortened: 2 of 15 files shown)


lwgray:
Both P1s confirmed. Working on fixes now. Will post a detailed response when both land in a follow-up commit.

lwgray:
## Both Codex P1s addressed (commit 4980eb5)

### P1 #1 — Non-RuntimeError exceptions escaping the fallback

**Root cause**: \`decompose_by_contract\` could raise \`TypeError\`/\`AttributeError\` from malformed-but-parseable LLM JSON (\`estimated_minutes: null\` → \`float(None)\` → TypeError, non-string \`description\` → \`.strip()\` → AttributeError). \`generate_structured_response\` parses JSON but does not enforce the schema, so these leaked past the narrow \`(ValueError, RuntimeError)\` handler and crashed project creation instead of triggering the advertised feature-based fallback.

**Two-layer fix**:

**Layer 1** — the fallback handler now catches \`Exception\` (with type name in the log so the underlying failure is still visible):

\`\`\`python
except Exception as e:
    logger.warning(
        f\"[decomposer] contract_first decomposer failed \"
        f\"({type(e).__name__}): {e}; falling back to \"
        f\"feature_based\"
    )
    return None
\`\`\`

**Layer 2** — \`decompose_by_contract\` now coerces every LLM output field defensively so most of these never fire in the first place:

- \`estimated_minutes\`: \`float(raw) if raw is not None else 8.0\`, with \`try/except TypeError|ValueError\` fallback
- \`description\`: \`str(raw).strip() if raw is not None else \"\"\`
- \`name\`, \`responsibility\`, \`contract_file\`, \`provides\`, \`requires\`: all run through \`str(... or \"\")\` or equivalent
- Non-dict task (e.g. LLM returns a bare string in the \`tasks\` array) → clean \`RuntimeError\` with type name so the fallback fires instead of a confusing downstream error

New tests pin the behavior:

- \`test_malformed_estimated_minutes_coerced_to_default\`
- \`test_malformed_description_coerced_to_string\`
- \`test_non_dict_task_raises_runtime_error\`

### P1 #2 — \`Task.responsibility\` not persisted through kanban providers

**Root cause**: Verified against the files you cited. SQLite (\`src/integrations/providers/sqlite_kanban.py:1538-1566\`) persists \`source_context\` but NOT \`responsibility\` as a top-level column. Planka (\`src/integrations/kanban_client.py:938-952\`) persists neither — \`_card_to_task\` constructs \`Task\` with a minimal field set.

Result: the CONTRACT RESPONSIBILITY prompt layer silently skipped for tasks reloaded from the board, defeating the whole point of PR 2. Confirmed this would have failed Experiment 4 in production.

**Three-layer fix**:

1. **source_context redundancy**: \`decompose_by_contract\` now stores \`responsibility\` in \`source_context\` in addition to the top-level field. SQLite's source_context round-trip carries it automatically.

2. **Description marker**: \`decompose_by_contract\` embeds a structured HTML comment marker in \`task.description\`:

   \`\`\`
   
   \`\`\`

   Description is the one field every kanban provider round-trips as core data, so this marker survives even Planka.

3. **\`_parse_contract_metadata\` helper**: new in \`src/marcus_mcp/tools/task.py\`. Looks in three places in priority order:
   1. \`Task.responsibility\` + \`source_context[\"contract_file\"]\` (fresh decomposer output)
   2. \`source_context[\"responsibility\"]\` (SQLite round-trip survivor)
   3. Parsed from the description marker (Planka survivor)

   \`build_tiered_instructions\` Layer 1.3 and \`ai_analysis_engine.generate_task_instructions\` both use the helper. Layer fires for any of the three paths.

New tests pin the persistence fallback:

- \`test_sqlite_reload_source_context_path\` — SQLite provider simulation
- \`test_planka_reload_description_marker_path\` — Planka provider simulation
- \`test_no_marker_and_no_source_context_no_layer\` — legacy task stays legacy
- \`test_malformed_marker_ignored\` — missing closing \`-->\` → no half-surfaced garbage
- \`test_priority_order_direct_responsibility_wins\` — precedence pinned
- \`test_responsibility_persisted_in_source_context\` — decomposer output verified
- \`test_description_embeds_marker_for_persistence_fallback\` — marker format verified

### Test results

- \`pytest -m unit\`: **435/435 passed** (391 baseline + 44 new across PR 2b + Codex fixes)
- mypy clean on all touched files

### Follow-up

Added a note in the commit message: consider adding \`Task.responsibility\` as a first-class column in \`sqlite_kanban\` and Planka persistence to drop the description marker fallback. For now, the marker is the universal round-trip path and the two fresher paths (direct field, source_context) exist as priority 1 and 2 for providers that support them. Will file a follow-up issue if you want to track it separately.

Ready for re-review.

PR #329, opened 2026-04-11
test(#320): contract-first smoke test harness

lwgray:
## Summary

Standalone smoke-test script that validates the GH-320 PR 2 contract-first decomposition pipeline end-to-end against a running Marcus HTTP server. Reusable as a regression harness for any future PR touching the contract-first path.

Ran against develop after #327 merge. **All 4 invariants pass.**

## What it checks

| Invariant | What it proves | How it checks |
|---|---|---|
| 1 | Domain discovery produces ≥2 distinct domains | Counts unique slug prefixes under \`project_root/docs/\` |
| 2 | Implementation tasks carry contract-first metadata | Reads SQLite kanban directly, resolves via 3-tier fallback (direct → source_context → description marker) |
| 3 | Integration task has CONTRACT-FIRST authoritative preamble | Scans integration task description for marker text |
| 4 | Agent prompt contains CONTRACT RESPONSIBILITY layer | Registers test agent, calls \`request_next_task\`, greps the returned 5.7k-char instructions |

Invariant 4 is the end-to-end closure for Codex P1 from #327: invariants 2 and 3 prove data exists in state, invariant 4 proves it reaches the agent through \`build_tiered_instructions\` Layer 1.3.

## Usage

\`\`\`bash
# Prereq: Marcus HTTP server running at localhost:4298
python dev-tools/examples/gh320_experiment_4_smoke.py
# Exit 0 = all invariants pass
# Exit 1 = regression in contract-first pipeline
\`\`\`

## Test results on develop (2026-04-11, post-#327)

\`\`\`
🎉 All invariants hold. Pipeline is wired correctly.
  ✅ Domain count >= 2                                  (3 domains found)
  ✅ Impl tasks have responsibility                      (3/3 via source_context path)
  ✅ Integration task has preamble
  ✅ Agent prompt contains CONTRACT RESPONSIBILITY       (via request_next_task)
\`\`\`

## Pre-existing bugs surfaced (NOT blocking this PR)

Running the smoke test revealed two issues that live outside PR 2's scope:

1. **Contract scope bug** — \`_generate_contracts_by_domain\` produces full-system contracts per domain instead of domain-scoped ones. The three \`interface-contracts.md\` files have contradictory field types for shared entities (e.g. \`time.lastUpdated\` is typed as ISO string in one file, number in another, Date in a third). Lives in PR #322's contract-generation prompt template. **Blocks the real Experiment 4 until fixed** but not this PR.

2. **Observability bug in \`nlp_tools.py:575\`** — logs \`Task type breakdown: {'unknown': N}\` on every project (feature-based and contract-first) because \`getattr(task, 'task_type', 'unknown')\` references a field that doesn't exist on the Task model. Low urgency.

Both will get their own issues filed post-Experiment 4.

## Design notes

- **SQLite direct reads**: \`query_project_history\` returns historical tasks; fresh projects have none. The smoke test reads the live kanban directly via sqlite3 for task state. This is acceptable because the smoke test is a localhost dev tool, not a production path.
- **Priority-2 responsibility resolution is expected**: SQLite schema has no \`responsibility\` column, so the direct \`Task.responsibility\` field can never survive round-trip. \`source_context\` redundancy catches it. This is Codex P1 fallback design working as intended.
- **No config file required**: the script uses \`project_root\` in options so artifacts land in a fresh \`/tmp/gh320_smoke_<timestamp>\` directory per run — doesn't pollute repo tree or interfere with Larry's production Marcus config.

## Test plan

- [x] mypy clean on the new file
- [x] flake8 clean
- [x] pydocstyle clean
- [x] Runs end-to-end against live Marcus HTTP server and reports all 4 invariants
- [x] Exit code 0 on pass, 1 on failure
- [ ] Post-merge: use this harness before running the full Experiment 4 (2-agent run)

### DIFF
--- a/dev-tools/examples/gh320_experiment_4_smoke.py
+++ b/dev-tools/examples/gh320_experiment_4_smoke.py
@@ -1,0 +1,723 @@
+#!/usr/bin/env python3
+"""
+GH-320 Experiment 4 — contract-first decomposition smoke test.
+
+Connects to a running Marcus HTTP server, creates a project with
+``MARCUS_DECOMPOSER=contract_first`` enabled via the options dict,
+waits for Phase A (background design content + contract generation)
+to complete, and verifies the contract-first invariants:
+
+1. More than one domain was discovered (single-domain PRDs collapse
+   to the same topology-coupling failure mode contract-first is
+   supposed to solve — the Experiment is vacuous on a 1-domain
+   project).
+2. Implementation tasks carry ``Task.responsibility`` set (or have
+   the ``MARCUS_CONTRACT_FIRST`` marker embedded in their
+   description as the persistence fallback).
+3. The integration verification task's description includes the
+   ``CONTRACT-FIRST PROJECT`` preamble naming the contract file.
+
+This is a pipeline validation gate. It does NOT spawn agents or
+measure contribution distribution — that is the full Experiment 4
+and requires multi-agent orchestration outside this script.
+
+If all three invariants hold, PR #327's contract-first path is
+wired correctly end-to-end and the real Experiment 4 (2-agent
+snake or dashboard run) can proceed. If any invariant fails, this
+script reports which one and PR #327 needs a fix before burning
+agent API credits.
+
+Usage
+-----
+Prerequisites:
+    - Marcus HTTP server running (``python -m src.marcus_mcp.server --http``)
+    - Default endpoint: ``http://localhost:4298/mcp``
+    - A writable ``project_root`` for the test project (defaults to
+      ``/tmp/gh320_experiment_4_smoke_<timestamp>``)
+
+Run:
+    python dev-tools/examples/gh320_experiment_4_smoke.py
+"""
+
+from __future__ import annotations
+
+import asyncio
+import json
+import re
+import sqlite3
+import sys
+import time
+from pathlib import Path
+from typing import Any, Dict, List, Optional
+
+# Add project root to sys.path so imports resolve when running this
+# script directly from the dev-tools/examples directory.
+project_root = Path(__file__).parent.parent.parent
+sys.path.insert(0, str(project_root))
+
+from src.worker.inspector import Inspector  # noqa: E402
+
+# --------------------------------------------------------------------------
+# Configuration
+# --------------------------------------------------------------------------
+
+MARCUS_URL = "http://localhost:4298/mcp"
+PROJECT_NAME = f"gh320_smoke_{int(time.time())}"
+PROJECT_ROOT = f"/tmp/{PROJECT_NAME}"  # nosec B108 — smoke test temp dir
+
+# Rich description that forces at least 2 domains.
+#
+# The dashboard-v16 experiment proved feature-based decomposition
+# could split Weather and Time cleanly when the PRD described them
+# as independent concerns. The prior smoke test (2026-04-11)
+# collapsed "weather + time widgets" into one domain because the
+# description was too terse. This version is explicit about
+# ownership boundaries and data-flow independence.
+PROJECT_DESCRIPTION = (
+    "Build a dashboard web app with two independent widgets that "
+    "share no state or code.\n\n"
+    "1. WeatherWidget: fetches current weather from a public API "
+    "(e.g. OpenWeatherMap), displays temperature and conditions, "
+    "refreshes every 10 minutes. Owns all weather-related data "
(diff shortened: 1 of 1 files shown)


lwgray:
## Invariant 5 added (commit f9433ec2)

Added cross-file type contradiction detection as the 5th smoke test invariant. The regex + normalizer details are in the commit message.

**This invariant was the one that caught the contract-scope bug during pre-flight testing.** Before the prompt fix landed in PR #330, this check would flag real contradictions like \`lastUpdated: number (Weather file) vs string, ISO 8601 (Time file)\`. After PR #330's scope clamp, it passes on consecutive fresh LLM runs.

**This PR (#329) and PR #330 should merge together** — they solve the same problem from two sides:
- #329: the regression guard that catches scope leaks when they happen
- #330: the prompt fix that prevents them in the first place

Without #330, invariant 5 fails on every run. Without #329, a future refactor could re-introduce the scope leak silently and no one would notice until Experiment 4 (or production) drifted. They're a matched pair.

lwgray:
## Review response (commit 8877bb8a)

Addressed both P1s and the P2 cleanups.

### P1.1 — Project selection race

You were right — \`create_project\` is supposed to auto-select but that's best-effort behavior. If it fails, \`request_next_task\` could pull a task from a DIFFERENT project and invariant 4 would evaluate the wrong data, silently passing (or failing) on something unrelated.

Fix: explicit \`select_project\` call with the just-created \`project_id\` before \`register_agent\` + \`request_next_task\`. Closes the race at the call site.

\`\`\`python
await session.call_tool(
    \"select_project\",
    arguments={\"project_id\": project_id},
)
\`\`\`

### P1.2 — Contract detection restricted to CONTRACT RESPONSIBILITY section

Implemented exactly as you suggested. Previous check scanned the full 5700+ char instructions text for \`\"Contract file:\"\` or \`\".md\"\` — other layers in \`build_tiered_instructions\` (workflow, dependency awareness, README generation) can produce false positives.

Fix: extract the CONTRACT RESPONSIBILITY section by finding its header and walking forward to the next major section separator (the emoji separators Marcus uses plus \`\\n\\n## \`), then check contract-file + read-directive presence only within that slice. I also tightened the contract-file match from generic \`\".md\"\` to the specific contract filenames: \`interface-contracts\`, \`-contracts.md\`, \`api-contracts\`, \`data-models\`, \`architecture.md\`.

The section-extraction walks these separators since Marcus uses emoji-prefixed layer headers:

\`\`\`python
for separator in (\"\\n\\n📚\", \"\\n\\n🔗\", \"\\n\\n📝\", \"\\n\\n⚠️\", \"\\n\\n## \"):
    sep_idx = instructions_text.find(separator, section_floor)
    if sep_idx != -1 and sep_idx < end:
        end = sep_idx
\`\`\`

### P2 — Module-level regex + named constants

Done. Hoisted \`MARKER_PATTERN\` and \`TYPE_ANNOTATION_PATTERN\` to module level. Added three named constants replacing magic numbers:

\`\`\`python
PHASE_A_TIMEOUT_SECONDS = 120
PHASE_A_COMPLETION_BUFFER_SECONDS = 10
PHASE_A_POLL_INTERVAL_SECONDS = 5
\`\`\`

### P3 — Phase A timeout error message

Now includes the specific reason (\"artifact dir does not exist\" vs \"no .md files under artifact dir\") instead of a generic \"timed out\" message.

### Not done (with rationale)

**Path traversal validation on PROJECT_NAME** — skipped. This is a localhost dev tool running under the user's own identity against their own \`/tmp\`. Defense-in-depth has little value for a script the user can already \`rm -rf\` if they want to.

**File stability checks for Phase A completion** — skipped. The 10s buffer after first artifact is pragmatic and the smoke test has run 7 times now without an empty-file-read failure. Adding stability checks if it becomes flaky; not before.

### Verification

All 5 invariants still pass end-to-end against live Marcus after the fixes:

\`\`\`
🎉 All invariants hold. Pipeline is wired correctly.
  ✅ Domain count >= 2
  ✅ Impl tasks have responsibility
  ✅ Integration task has preamble
  ✅ Agent prompt contains CONTRACT RESPONSIBILITY   (tighter check now)
  ✅ No cross-file type contradictions
\`\`\`

The tighter Invariant 4 check still passes — confirms the CONTRACT RESPONSIBILITY section really does contain both \`Contract file:\` and the specific contract filename, not just similar strings elsewhere in the 5800-char instruction text.

Ready for re-review.

PR #330, opened 2026-04-11
fix(#320): tighten contract-generation prompt scope clamp

lwgray:
## Summary

Fixes a cross-domain field leakage bug in per-domain contract generation that would have silently corrupted Experiment 4. LLM generated \"per-domain\" contracts were actually full-system contracts with contradictory type definitions for shared fields across files. Scope clamp + forbidden-patterns list in the prompts eliminates it.

**Evidence**: smoke test via PR #329 now passes all 5 invariants including cross-file type consistency.

## Problem

Phase A's \`_generate_contracts_by_domain\` (shipped in PR #322) runs the same prompt template per domain with only the domain name swapped. When tested on a multi-domain project, the LLM consistently produced full-system contracts per domain instead of domain-scoped ones.

Three \`interface-contracts.md\` files were generated for a Weather + Time + Dashboard project. All three described the entire system, and they had **contradictory type definitions for shared fields**:

| Field | Weather file | Time file | Dashboard file |
|---|---|---|---|
| \`time.lastUpdated\` | number (Unix) | string (ISO 8601) | ISO 8601 string |

Agents assigned one file each as \"authoritative\" would produce incompatible code at integration time — the exact silent-drift failure mode contract-first decomposition is supposed to prevent.

## Root cause

The original prompt says \"describe boundaries and contracts for this design\" but never tells the LLM to scope its output to a specific domain. The LLM reads the full project description and the domain task description, then writes a full-system view because it's the most natural thing to produce.

## Fix

Three prompt template changes in \`src/integrations/nlp_tools.py\`:

### 1. \`_ARTIFACT_PROMPT\` — add CRITICAL SCOPE CONSTRAINT section

Explicitly names the specific \`{domain_name}\` and lists:
- **Include**: fields owned by this domain
- **Exclude**: fields owned by other domains (reference by name only)

### 2. \`_INTERFACE_CONTRACTS_PROMPT\` — add SCOPE CONSTRAINT + FORBIDDEN PATTERNS

Beyond the scope clamp, this prompt also includes a specific forbidden-patterns section naming the failure modes observed in the smoke test:

> **FORBIDDEN PATTERNS (must not appear in this document)**
>
> 1. Tables that include fields owned by other domains
> 2. Re-declaring another domain's field with type information
> 3. Describing another domain's props interface
> 4. \"For reference\" or \"for clarity\" type definitions of external fields

With concrete good/bad examples for each.

### 3. \`_DECISIONS_PROMPT\` — add lighter scope constraint

So cross-domain decisions don't get duplicated across per-domain decision lists.

### 4. Plumbing

Add \`{domain_name}\` parameter to all three prompt \`.format()\` calls in \`_generate_single_artifact\` and \`_generate_single_decisions\` so the templates can interpolate it.

## Validation

Validated via \`dev-tools/examples/gh320_experiment_4_smoke.py\` (PR #329). Iteration history:

| Run | Change | Contradictions |
|---|---|---|
| 0 | Pre-fix | Many (multiple real type mismatches) |
| 1 | Initial scope clamp | 7 (mostly regex false positives + 1 real) |
| 2 | Tightened regex | 1 (\`apikey\` string vs string+required — stylistic) |
| 3 | Better normalizer | 2 (\`error\`, \`timezone\` — stylistic syntax differences) |
| 4 | Better normalizer v2 | 1 real (\`lastUpdated\` leaked into Time's summary table) |
| 5 | **Added FORBIDDEN PATTERNS section** | **0** |
| 6 | Reproducibility check | **0** |

**Two consecutive clean runs on fresh LLM calls prove the scope clamp + forbidden patterns list works.**

## Kaia's Option A succeeded

Per Kaia's pre-flight vote: \"Option A (tighten _ARTIFACT_PROMPT and _INTERFACE_CONTRACTS_PROMPT with explicit scope clamp) with Option C (redesign to single shared contract file) as fallback if A fails empirically.\"

The data says A worked within the 2-iteration budget Kaia predicted. Option C (single shared contract file refactor) is not needed.

## Existing tests

All 19 existing tests in \`tests/unit/integrations/test_design_autocomplete.py\` still pass. The prompt templates changed but the contract-generation code paths are identical.

## Test plan

- [x] 19/19 existing design_autocomplete unit tests pass
- [x] Smoke test invariant 5 passes on two consecutive fresh LLM runs
- [x] mypy clean on \`src/integrations/nlp_tools.py\`
- [x] Scope clamp logic also reduces total field count in generated files (fewer duplicated fields across files → smaller, more focused contracts)
- [ ] **Post-merge**: run the actual Experiment 4 (2-agent run) now that contracts are trustworthy

## Related

- #320 (contract-first task decomposition)
- #322 (domain-keyed contract generation, PR 1 — shipped the prompt templates this fixes)
- #327 (contract-aware decomposer, PR 2 — merged)
- #329 (smoke test harness that caught this bug — must land with this to keep regression guard)

### DIFF
--- a/src/integrations/nlp_tools.py
+++ b/src/integrations/nlp_tools.py
@@ -1514,14 +1514,50 @@ class NaturalLanguageFeatureAdder(NaturalLanguageTaskCreator):
 _ARTIFACT_PROMPT = """\
 You are a senior software architect working on: {project_name}
 
-## Project Description
+## Project Description (for context only)
 {project_description}
 
 ## Your Design Task
 {task_description}
 
 ## Your Current Assignment
-Generate the {artifact_label} document for this design.
+Generate the {artifact_label} document for the **{domain_name}** domain \
+specifically.
+
+## CRITICAL SCOPE CONSTRAINT (GH-320)
+
+This document describes ONLY the **{domain_name}** domain. You are one of \
+several architects producing contracts for this project, each scoped to \
+a single domain. If you describe fields, types, or interfaces owned by \
+another domain, those other architects will produce their own definitions \
+for the same thing, and the definitions will contradict each other at \
+integration time.
+
+**What to include**: fields, types, identifiers, configuration values, \
+API endpoints, and interfaces OWNED BY the {domain_name} domain. Things \
+the {domain_name} module produces or stores internally.
+
+**What to exclude**: fields, types, or interfaces owned by other domains. \
+If the {domain_name} domain must interact with another domain's data, \
+reference that data BY NAME ONLY and point to the other domain's contract \
+file. Never redefine another domain's fields, never guess at their types, \
+never include their interface definitions in this file.
+
+**Good cross-domain reference**:
+> "When a user selects a timezone, the {domain_name} domain receives a \
+> TimezoneIdentifier from the TimeWidget domain (see time-widget-system \
+> contracts for its shape)."
+
+**Bad cross-domain leak**:
+> "TimezoneIdentifier (string, IANA format, e.g., 'America/New_York') — \
+> defined by TimeWidget domain."
+> (Do NOT redefine another domain's types. Refer to them by name only.)
+
+**Test before you write each field**: "Is this field owned by the \
+{domain_name} domain, or by another domain?" Only describe fields where \
+the answer is **{domain_name}**.
+
+## Content Guidelines
 
 Describe WHAT each component does and HOW components connect to each \
 other. Focus on behavior, responsibilities, data flow, and integration \
@@ -1532,12 +1568,14 @@ names, or internal implementation details. The implementing developer \
 decides those. Your job is to define the WHAT and WHY, not the HOW.
 
 However, you MUST be concrete and specific about any identifier, name, \
-or value that will be shared across module boundaries — field names in \
-data models, storage keys, event names, environment variable names, \
-port numbers, API response shapes, and status/enum values. When \
-multiple modules must agree on a name or value to interoperate, that \
-name or value is an interface contract, not an implementation detail. \
-State it explicitly.
+or value that the {domain_name} domain owns and that will be shared \
+across module boundaries — field names in data models, storage keys, \
+event names, environment variable names, port numbers, API response \
+shapes, and status/enum values. When multiple modules must agree on a \
+name or value to interoperate, that name or value is an interface \
+contract, not an implementation detail. State it explicitly — **but \
+only if the {domain_name} domain is the owner/producer of that \
+identifier**.
 
 Good: "The time display updates every second using the browser's \
 Date API and supports timezone conversion."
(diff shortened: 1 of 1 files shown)


lwgray:
## Review response (commit bee2b5b9)

### Codex P2 — backslash line continuations mangling blockquote examples

**This was a real, important bug.** Codex was right. The prompt templates had \`\\\` line continuations inside the Good cross-domain reference examples, which Python's string parser joined into one line. Rendered output was:

\`\`\`
> "The Weather domain receives a TimeEntity as input from > the TimeWidget domain. See contracts > for the authoritative ..."
\`\`\`

Orphan \`>\` characters mid-sentence instead of a three-line blockquote. **Exactly the opposite of what the example was trying to teach** — the prompt instructs the LLM to produce well-formed blockquotes while presenting it with a malformed one.

The Bad examples rendered correctly because they didn't use \`\\\` line continuations. Only the most important teaching examples (the Good ones) were broken. This would measurably degrade the prompt's effectiveness, and probably explains why my first scope-clamp iteration only partially worked — the LLM was learning from a bad example.

**Fix**: removed \`\\\` continuations from the Good/Bad example blockquotes in both \`_ARTIFACT_PROMPT\` and \`_INTERFACE_CONTRACTS_PROMPT\`. Also fixed the numbered FORBIDDEN PATTERNS items which had \`{domain_name}- \\n   owned fields\` that rendered as \`Weather-   owned fields\`. Kept \`\\\` for prose lines that should genuinely flow as one output line.

**Verification**: rendered both prompts with dummy parameters and confirmed the Example sections now show proper multi-line blockquotes with \`>\` on each line.

### Claude-bot P1 — domain_name template injection defense

Added defensive validation in both \`_generate_single_artifact\` and \`_generate_single_decisions\`:

\`\`\`python
if not domain_name or not domain_name.strip():
    raise ValueError(...)
if \"{\" in domain_name or \"}\" in domain_name:
    raise ValueError(...)
\`\`\`

The domain name flows into \`.format()\`, so curly braces in it would either raise KeyError mid-request or silently corrupt the prompt. Blocking at the call site with a clear error.

The threat model here is a future Marcus code path or prompt drift, not an external attacker — the domain name comes from Marcus's internal domain discovery. But cheap, clean, defensive, so done.

### Claude-bot P1 — regression test for cross-domain type consistency

**Already covered by Invariant 5 in the smoke test (PR #329).** The smoke test script IS the regression test — it's specifically designed to catch cross-domain type contradictions. Running it two consecutive times with the scope-clamped prompts produced 0 contradictions both times. Adding a unit test that mocks the LLM output would test less than what the smoke test already proves end-to-end.

### Not done (with rationale)

- **Unit tests for prompt template parameter substitution**: low value. \`.format()\` is a Python stdlib guarantee. The new validation covers the one non-trivial concern (braces in domain_name).
- **Performance monitoring of longer prompts**: not actionable in code. Phase A parallelism (#304) already caps concurrent LLM calls.
- **Post-processing scope validation**: would be defense-in-depth but Invariant 5 in the smoke test already serves this role at the verification layer rather than the generation layer.

### Verification

- 19/19 existing \`test_design_autocomplete.py\` unit tests pass
- mypy clean on \`src/integrations/nlp_tools.py\`
- Both prompt templates render the Example blockquotes with correct three-line structure (verified by rendering with dummy parameters)

Ready for re-review. **This is also the second time Codex has caught a real bug in my PR that Claude-bot missed** — the line-continuation issue was subtle enough that a logic-level review could have easily skipped it. The combination of human-style review + syntactic review has been valuable.

Issue #388, opened 2026-04-22
bug: agents from concurrent experiments steal tasks across project boundaries


lwgray:
## Summary

Marcus is a singleton MCP server. All experiments share the same server instance. When `create_project` runs, it sets `kanban_client.project_id` on the server to the newly created project. Any agent from a *previously running* experiment that is still polling `request_next_task` will now receive tasks from the NEW project — because they share the same server context and `get_available_tasks` scopes by `kanban_client.project_id`, which just changed.

## Root Cause

`register_agent` takes no `project_id` parameter. There is no binding between an agent and the project it belongs to. `request_next_task` scopes task queries using `kanban_client.project_id` (a server-level field set by the most recent `create_project` call), not by agent identity.

## Evidence

Discovered during dashboard-v88 post-mortem investigation. Three tasks were attributed to `agent_unicorn_3` in the dashboard-v88 kanban DB:

- `154c48e0` — Design System Setup
- `b27a116b` — Implement Animated Digital Clock  
- `ec6884f4` — Shared Component Foundation

UD3's JSONL conversation log (dashboard-v88) showed only ONE successful `request_next_task` response — for `ec6884f4`. The other two task IDs were found in **dashboard-v87's `agent_unicorn_3` JSONL**, with full `report_task_progress` calls. v87's agents were still polling when v88's project was created, grabbed v88 tasks, and completed them — leaving v88's UD3 with nothing to do and the audit incorrectly attributing false completions to v88's UD3.

Similarly, `504bd791` (Tech Foundation Setup) was attributed to `agent_unicorn_1` in dashboard-v88's DB, but the actual `report_task_progress` calls were in **dashboard-v86's `agent_unicorn_1` JSONL**.

## Fix

**`register_agent`** — add required `project_id` parameter. Store `agent_id → project_id` mapping in server state.

**`request_next_task`** — look up the calling agent's registered `project_id`, pass it explicitly to `get_available_tasks` rather than relying on the server-level `kanban_client.project_id`.

**`spawn_agents.py` worker prompt** — update step 2 (register_agent) to read `project_id` from `project_info.json` and pass it to `register_agent`. All three Marcus access routes already have `project_id` available:
- Direct MCP: returned in `create_project` response
- Runners + Posidonius: both go through `spawn_agents.py` → `project_info.json`

Making `project_id` **required** (not optional) is intentional — optional would allow the contamination to silently continue for callers who forget to pass it.

## Impact

[...]

PR #392, opened 2026-04-22
fix(#388): atomic registration + wire project_id through stdio handler

lwgray:
## Summary

Follow-up to #390 addressing two Codex review findings.

**P2 — ghost state on rejected registration** (`agent.py`):
`state.agent_status[agent_id]` was written before the `project_id` guard, leaving a stale `WorkerStatus` entry in server state whenever registration was rejected. Moved the `project_id` validation to the top of the try block — registration is now atomic (nothing written on failure).

**P1 — stdio call path never forwarded project_id** (`handlers.py`):
The `register_agent` dispatch in the handlers extracted `agent_id/name/role/skills` but not `project_id`. Every registration via the stdio tool path returned `success=False`, silently blocking all non-runner agent startups. Fixed by:
- Extracting `project_id` from arguments and forwarding to `register_agent()`
- Declaring `project_id` in the tool's `inputSchema` properties
- Adding `project_id` to the schema's `required` list

## Test plan

- [ ] `pytest -m unit tests/unit/marcus_mcp/test_project_scoped_registration.py` — 14 tests, all pass

### DIFF
--- a/src/marcus_mcp/handlers.py
+++ b/src/marcus_mcp/handlers.py
@@ -152,8 +152,16 @@ def get_tool_definitions(role: str = "agent") -> List[types.Tool]:
                         "description": "List of agent's skills",
                         "default": [],
                     },
+                    "project_id": {
+                        "type": "string",
+                        "description": (
+                            "Project this agent is working on (required). "
+                            "Scopes task assignment to prevent cross-experiment "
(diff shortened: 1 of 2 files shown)


lwgray:
Closing — fixes will be pushed to #390 instead.

PR #450, opened 2026-04-28
fix(epictetus): emit phase progress events to epictetus_progress.jsonl

lwgray:
## Summary

- Add `_phase_done` helper to Epictetus SKILL.md, called at end of each phase
- Helper writes phase status, duration, and elapsed time to `../epictetus_progress.jsonl` (run_dir, parent of implementation/)
- Documents the full 13-phase structure (11 always-on + 2 conditional on `--session`)
- Posidonius tails this file to show live audit progress and estimate time remaining

## Why

Posidonius UI now has an Epictetus status indicator (CSS phase pips, `fetchEpictetusStatus()` JS, `/api/experiments/{name}/events` endpoint), but Epictetus itself was emitting no progress signal. Operators saw "audit running" with no indication of which phase or how long left. This change adds the missing signal at the source.

## Test plan

- [ ] Run an Epictetus audit; verify `epictetus_progress.jsonl` is written incrementally
- [ ] Confirm Posidonius UI updates phase pips as the file grows
- [ ] Verify audit completes with all 13 phases logged
- [ ] Confirm `--session`-only phases (2.5, 7.5) are skipped cleanly when flag absent

### DIFF
--- a/skills/epictetus/SKILL.md
+++ b/skills/epictetus/SKILL.md
@@ -38,6 +38,57 @@ every finding. You never guess — you verify.
   collection (Phase 2.5) and agent interrogation (Phase 7.5). Without this flag,
   Epictetus performs a code-only audit as before.
 
+## Progress Tracking
+
+**Define this helper once at the very start of execution, before Phase 1:**
+
+```bash
+PROGRESS_FILE="../epictetus_progress.jsonl"
+PHASE_START=$(date -u +%s)
+AUDIT_START=$(date -u +%s)
+# Clear any stale progress from a previous run
+> "$PROGRESS_FILE"
+
+_phase_done() {
+  local phase_num="$1" phase_name="$2" skipped="${3:-false}"
+  local now=$(date -u +%s)
+  local duration=$((now - PHASE_START))
+  local elapsed=$((now - AUDIT_START))
+  local status="complete"
+  [ "$skipped" = "true" ] && status="skipped"
+  printf '{"phase":%s,"name":"%s","status":"%s","completed_at":"%s","duration_seconds":%d,"elapsed_seconds":%d}\n' \
+    "$phase_num" "$phase_name" "$status" \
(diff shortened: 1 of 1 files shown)


lwgray:
Addressed Codex P1 in commit 3c705b1a:

- **Phase count corrected**: 14 total (10 always-on + 4 conditional), was 13 (11 + 2)
- **Phase 7.8 added to the table** — it was emitted in code but missing from metadata; this was the root inconsistency
- **Phase 7.7 conditionality fixed**: `--session OR kanban.db`, not "always" (matches text at line 485)
- **Skip-emit lines added** to Phase 7.7 and 7.8, matching the pattern already used by 2.5 and 7.5 — ensures the JSONL stream contains a record for every conditional phase, skipped or not
- **Note added** clarifying that consumers computing percent-complete or ETA must subtract skipped (`true`) phases from the denominator

Phase emission set is now self-consistent: every `_phase_done` call in the file corresponds to a row in the table, and every conditional phase has both a "complete" and "skipped" emit pattern documented.

lwgray:
Addressed Codex P2 in commit 35c0f490:

Two guards added before the INSERT:

1. **`os.path.isfile(db_path)`** before `sqlite3.connect` — matches the documented "skip silently if the file does not exist" intent at line 746. Prevents `sqlite3.connect` from auto-creating an empty DB at a stale path.

2. **`sqlite_master` query for the `persistence` table** after connecting — protects against a real Marcus DB that predates this schema. Avoids `no such table: persistence` failures.

Both failure modes now print a skip message and `sys.exit(0)`, preserving Phase 8.5's best-effort contract. The block now correctly handles three skip cases: unset MARCUS_DB, missing file, and missing schema.

lwgray:
Closed via merged PR. Manual close because PRs merge to `develop` not `main` (default branch); auto-close didn't fire. Will sync develop → main on v0.3.6.post2 release.

PR #452, opened 2026-04-28
feat(parallel): asyncio.Lock for kanban_client init — fixes 720s stall in batch experiments

lwgray:
## Summary

- Add module-level `_kanban_init_lock: asyncio.Lock` in `src/marcus_mcp/tools/nlp.py`
- Wrap kanban_client provider check + `KanbanFactory.create` + `connect()` in `async with _kanban_init_lock`
- Add `TestKanbanInitLock` test class with structural guard and serialization assertion (2 new tests, 14 total in `test_parallel_sqlite.py`)

## Why

In batch parallel-experiment runs, multiple `create_project` calls fire simultaneously against a single Marcus instance. Without serialization:

1. All concurrent calls evaluate `need_new_client = True` at the same time
2. They race to call `KanbanFactory.create()` and overwrite `state.kanban_client`
3. The last writer wins; earlier callers retain references to orphaned, disconnected clients
4. The dedup guard then detects the stalled call as "still running" and loops, producing the ~720s stall observed in batch runs

Wrapping the provider check + factory + connect in an `asyncio.Lock` serializes the kanban_client mutation. Concurrent calls now wait for the first to complete; subsequent callers see `current_provider == requested_provider` and skip re-init.

## Test plan

- [x] `pytest tests/unit/mcp/test_parallel_sqlite.py -v` — 14/14 pass
- [x] `mypy --strict src/marcus_mcp/tools/nlp.py` — no issues
- [x] `black` and `isort` clean
- [x] `flake8` clean (E402 fixed by reordering import + lock declaration)
- [ ] Run a 3-instance batch experiment; confirm no 720s stall on concurrent project creation

## Bright Line / Multi-Agency Compliance

This change is internal coordination (Marcus serializing its own state mutation). It does not affect agents — they continue to self-select work via `request_next_task` and the board remains the only communication channel. ✅

### DIFF
--- a/src/marcus_mcp/tools/nlp.py
+++ b/src/marcus_mcp/tools/nlp.py
@@ -5,12 +5,12 @@ This module contains tools for natural language project/task creation:
 - add_feature: Add feature to existing project using natural language
 """
 
-import asyncio
 import logging
 from datetime import datetime, timezone
 from pathlib import Path
 from typing import Any, Dict, Optional
 
+from src.core.event_loop_utils import EventLoopLockManager
 from src.integrations.nlp_tools import add_feature_natural_language
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


lwgray:
Addressed Codex P1 in commit 41d98cda:

**Root issue**: A module-level `asyncio.Lock()` instantiated at import time binds to the first event loop that touches it. Marcus HTTP transport may run requests on different event loops, so a request on a second loop would either hang or raise `RuntimeError: <Lock> is bound to a different event loop` — the same class of stall this fix was meant to prevent.

**Fix**: Switched to `EventLoopLockManager` (already in `src/core/event_loop_utils.py` for exactly this purpose):

- `_kanban_init_lock` → `_kanban_init_lock_manager: EventLoopLockManager`
- `async with _kanban_init_lock` → `async with _kanban_init_lock_manager.get_lock()`
- `get_lock()` resolves at call time, returning a lock bound to the *current* running event loop (lazily-created and cached in a `WeakKeyDictionary` keyed by loop)

**New test** `test_distinct_locks_per_event_loop` runs the manager in two separate `asyncio.run()` invocations and asserts the returned lock objects have different ids — proves the per-loop scoping works and the cross-loop binding error can't occur.

15/15 tests pass; mypy --strict clean.

lwgray:
Closed via merged PR. Manual close because PRs merge to `develop` not `main` (default branch); auto-close didn't fire. Will sync develop → main on v0.3.6.post2 release.

Issue #449, opened 2026-04-28
feat: user-outcome coverage as intent fidelity gate (replaces project-type-specific guards)


lwgray:
## Problem

The decomposer treats the spec as a flat list of independent features. There is no representation of *what the user must be able to do* with the finished product. As a result, task graphs routinely fail to include the surface area required to make the union of features into a usable product.

This is not a visual-product problem. It is an **intent fidelity** gap that recurs across project types:

| Spec | Features extracted | Missing | Result |
|------|--------------------|---------|--------|
| "build a snake game" | state, movement, collision, food, score | rendering | invisible game (snake_game-v31, 2026-04-28) |
| "user management service" | user model, auth, hashing, sessions | HTTP endpoints | a library, not a service |
| "markdown linter CLI" | rule engine, AST, error reporter | argparse, file I/O | a library, not a tool |
| "analytics pipeline" | aggregations, transforms, reports | data ingestion | transforms with no input |
| "chat app" | message store, websockets, auth, UI | layer wiring | parts in isolation |

Each failure passes existing decomposition because feature filtering optimizes for "essential features" without checking that the union yields an end-to-end usable product.

## Evidence: why current mechanisms don't catch this

- `_generate_success_criteria()` in `src/ai/advanced/prd/advanced_parser.py:2678` produces audit-log fluff: `"All tests passing"`, `"Application deployed and accessible"`. Never enforced, never consulted during decomposition.
- Per-task `acceptance_criteria` (line 459, 488) is descriptive output, not a constraint.
- Hardcoded type guards are accumulating: `"frontend"` (3493), `"e-commerce"` (3504), `"UI/UX Designer"` (3387). Each new project type adds another band-aid.
- Integration Verification gate signs off on `tests pass + build OK + HTTP 200`. None of these prove the user can accomplish anything (cf. snake_game-v31: HTTP 200 served `<main id="game-mount">` with no rendered content).

## Proposed Fix: User-Outcome Coverage Check

Add a step before feature filtering and after task decomposition.

### Stage 1 — Extract user-visible outcomes from spec

After PRD analysis, generate a list of `UserOutcome` records:

```python
@dataclass
class UserOutcome:
    id: str                    # "outcome_play_game"
    action: str                # "user can play the snake game in their browser"
    success_signal: str        # "snake visibly moves on a board, food appears, score updates"
    scope: str                 # "in_scope" | "out_of_scope" (per spec language)
```

Constraints on extraction:
[...]

Issue #447, opened 2026-04-27
feat: surface planning phase (PRD analysis, domain discovery, synthesis) in Cato swim lanes


lwgray:
## Problem

Marcus Phase A (PRD analysis → domain discovery → contract generation → task synthesis) runs entirely before any task hits the kanban board. Cato swim lanes only show board-visible work, making the entire planning phase invisible to observers.

This creates a false picture: from Cato's perspective, agents appear to spring into existence with a full task list and no prior work. The most expensive and consequential part of a run — where Marcus decides what to build and how to structure it — is unobservable.

## Impact

- No visibility into how long planning took
- No way to audit WHY a particular task decomposition was chosen
- Debugging failures requires reading raw logs instead of Cato
- Research claim that Marcus is "fully observable" is undermined

## Current State

`EventLog` writes `pipeline_events.jsonl` for Posidonius-level events. Marcus itself emits no events during Phase A. `conversation_logger.py` captures agent-facing MCP calls but Phase A happens before agents start.

## Proposed Fix

Two-part:

**1. Marcus: emit Phase A events to a log stream**
During `create_project` / PRD parse / domain discovery / synthesis, write structured events to `logs/planning_events.jsonl`:
```json
{"timestamp": "...", "phase": "domain_discovery", "status": "started", "project_id": "..."}
{"timestamp": "...", "phase": "domain_discovery", "status": "completed", "domains": ["auth", "api", "ui"], "duration_ms": 4200}
{"timestamp": "...", "phase": "contract_generation", "status": "completed", "contracts": 3}
{"timestamp": "...", "phase": "task_synthesis", "status": "completed", "tasks_created": 23}
```

**2. Cato: render planning events as a synthetic "Planning" swim lane**
Read `planning_events.jsonl` alongside board tasks. Show Phase A as a collapsed swim lane above the agent lanes with phase pips (PRD → Domains → Contracts → Tasks).

## Acceptance Criteria

- [ ] Planning phase visible in Cato before first agent task appears
- [ ] Duration of each Phase A step shown
- [ ] Domain names and task counts surfaced
- [ ] Planning lane collapses when expanded view not needed

Issue #446, opened 2026-04-27
feat: foundation tasks must log_decision their public API surface so downstream consumers can discover via get_task_context


lwgray:
## Problem

When Marcus generates pre-fork foundation tasks (shared design system, API contracts, DB schema), task descriptions specify WHAT to build but not HOW downstream agents should consume the artifact.

v80 experiment demonstrated this failure mode: agents built a design system correctly but dependent agents wired it incorrectly because nothing in the task description said *how* consumers should import or reference it.

## Root Cause

`decomposer.py` creates foundation task descriptions using domain discovery output, but the LLM prompt asks for what to build, not for consumption contracts. `_generate_wiring_tasks` generates `provides`/`requires` pairs for integration wiring, but the foundation task description itself lacks:
- Import path conventions
- Interface contracts (function signatures, class APIs)
- Configuration keys downstream tasks must reference

## Impact

Agents that receive a downstream task have no way to know how to correctly consume the foundation artifact unless they happen to discover it from the code — which they often get wrong.

## Expected Behavior

Foundation task descriptions should include a `## Consumption Contract` section specifying:
```
## Consumption Contract
- Import: `from shared.design_system import theme, Button`
- Config key: `DESIGN_SYSTEM_VERSION` in `pyproject.toml`
- Required: call `theme.apply()` before rendering any component
```

## Proposed Fix

Extend the domain discovery prompt or foundation task synthesis step to generate a `consumption_contract` field alongside the `description`. Inject it into the task description before board creation.

## Evidence

- v80 experiment: design system built, dependent agents wired incorrectly
- Related: `src/marcus_mcp/coordinator/decomposer.py` `_decompose_with_contracts()` — contracts generated at decomposition time but not surfaced in task description body

lwgray:
Closed via merged PR. Manual close because PRs merge to `develop` not `main` (default branch); auto-close didn't fire. Will sync develop → main on v0.3.6.post2 release.

PR #453, opened 2026-04-28
feat(#449): user-outcome coverage foundation — extractor, coverage, gap-fill

lwgray:
Closes part of #449 — the foundational module layer.  Decomposer integration is the follow-up PR.

## Summary

Introduces three pure modules that turn the v31-snake failure mode into a measurable property:

- `src/ai/advanced/prd/outcome_extractor.py` — `UserOutcome` dataclass with strict `__post_init__` validation, plus `extract_user_outcomes(spec, llm_client)` to produce a list of outcomes from a spec via one LLM call.
- `src/marcus_mcp/coordinator/outcome_coverage.py` — `compute_coverage`, `find_gaps`, `compute_intent_fidelity_score`, `fill_gaps`.
- 30 new unit tests including a snake_game-v31 regression test that reproduces the production failure end-to-end with mocked LLMs.

## Why

snake_game-v31 audit (2026-04-28): the spec `"build a snake game"` produced a task graph with no rendering task because the decomposer treated rendering as implicit.  HTTP 200 + tests pass; user sees an empty `<main id="game-mount">`.

This pattern recurs across project types — APIs missing endpoint handlers, CLIs missing argparse, pipelines missing data ingestion.  The systemic fix is to treat user-visible outcomes as decomposition constraints, not optional metadata.  See #449 for the full failure-mode taxonomy.

### `UserOutcome` dataclass
Strict validation in `__post_init__`:
- `action` must contain `user can` / `user is able to` / `users can` / etc.
- Vague phrasings (`user can use the app`) rejected with a clear error
- `scope` must be `in_scope` or `out_of_scope`
- `id` and `success_signal` must be non-empty

### `extract_user_outcomes`
- One LLM call per spec, validated against the schema
- Out-of-scope outcomes are **retained** (not dropped) so the audit trail records what the LLM considered and rejected based on spec language (`"no auth"`, `"no accounts"`)
- Empty outcome lists raise — every spec implies at least one user-visible capability
- ID collisions raise — would silently corrupt the coverage map

### `outcome_coverage` module
- `compute_coverage(outcomes, tasks, mapper=None)` — outcome_id → list of task_ids that address it.  Default keyword-overlap mapper is suitable for unit tests; tests inject explicit mappers for deterministic behavior.  **Decomposer integration (next PR) will use an LLM-backed mapper** — the keyword default is provably inadequate for the snake-v31 case, where domain nouns (`snake`, `food`, `score`) overlap heavily between outcomes and internal logic tasks.  This is documented in the regression test.
- `find_gaps` returns only in-scope outcomes with no covering task.  Out-of-scope outcomes are retained for audit but excluded from gaps.
- `compute_intent_fidelity_score` = `covered_in_scope / total_in_scope`, vacuously `1.0` when there are no in-scope outcomes (alternative — NaN, 0, raise — surprises consumers).
- `fill_gaps(spec, gaps, llm_client)` — single LLM call producing replacement task dicts.  **No call when gaps is empty** — this is the cost invariant.

### Tests (30 new)
- `tests/unit/ai/test_outcome_extractor.py` — 12 tests covering dataclass invariants, extraction success, and every error path
- `tests/unit/coordinator/test_outcome_coverage.py` — 16 tests covering coverage, gaps, score, gap-fill
- `tests/unit/coordinator/test_intent_fidelity_pipeline.py` — 2 end-to-end tests reproducing the snake_game-v31 scenario:
  - `test_missing_render_task_is_detected_and_filled`: extracts outcome → simulates v31 logic-only task graph → coverage finds gap → gap-fill produces a render task → final fidelity is 1.0
  - `test_full_coverage_skips_gap_fill_call`: when the graph already covers all outcomes, no LLM call is made (cost invariant)

## What's NOT in this PR (deliberate)

- **Decomposer integration** — wiring `extract_user_outcomes` into `advanced_parser.py` and the coverage check into `decomposer.py`.  Those files are 4,500 and 1,100 lines respectively; integration deserves its own review and PR.
- **Feature flag `MARCUS_OUTCOME_COVERAGE`** — pairs with the integration above.
- **Cato surfacing of `intent_fidelity_score`** — covered by #447 scope.
- **Integration Verification using outcome `success_signal` as gate (Stage 5)** — explicitly deferred per #449 description.

## Bright Line / Multi-Agency Compliance

`UserOutcome.action` says **WHAT** the user must be able to do.  `success_signal` describes observable evidence.  Neither prescribes HOW.  Two agents given the outcome `"user can play the snake game"` can implement it with canvas, DOM, SVG, WebGL, or ASCII art in a terminal — all legitimately different.  Marcus says WHAT; agents pick HOW.  ✅ Coordination, not control.

## Test plan

- [x] `pytest tests/unit/ai/test_outcome_extractor.py tests/unit/coordinator/test_outcome_coverage.py tests/unit/coordinator/test_intent_fidelity_pipeline.py` — 30/30 pass
- [x] `mypy --strict src/ai/advanced/prd/outcome_extractor.py src/marcus_mcp/coordinator/outcome_coverage.py` — no issues
- [x] black / isort / flake8 / pydocstyle clean
- [ ] Full `pytest -m unit` — verify no regressions in the broader unit suite

## Related

- Closes part of #449 (foundational layer); decomposer integration to follow
- Builds on #446 (consumption contracts) and #447 (Cato planning visibility) — those address HOW tasks describe interfaces and observability of planning; this addresses WHICH tasks exist.

### DIFF
--- a/src/ai/advanced/prd/outcome_extractor.py
+++ b/src/ai/advanced/prd/outcome_extractor.py
@@ -26,14 +26,28 @@ from typing import Any, List
 
 from src.utils.json_parser import parse_ai_json_response
 
-# Phrases that mark a real user-capability action.  At least one must
-# appear (case-insensitive) in :attr:`UserOutcome.action`.
-_USER_CAPABILITY_PHRASES: tuple[str, ...] = (
-    "user can",
-    "user is able to",
-    "user must be able to",
-    "users can",
-    "users are able to",
+# Patterns that mark a real user-capability action.  Word-boundary
+# regex (not substring) so that ``"user cannot ..."`` and ``"user
+# can't ..."`` do NOT match — those are negations and contradict the
+# positive-capability invariant.  The lookahead ``(?!')`` rejects the
+# ``"can't"`` contraction.
+_USER_CAPABILITY_PATTERNS: tuple["re.Pattern[str]", ...] = (
+    re.compile(r"\busers? can\b(?!')", re.IGNORECASE),
+    re.compile(r"\busers? (?:is|are) able to\b", re.IGNORECASE),
+    re.compile(r"\busers? must be able to\b", re.IGNORECASE),
+)
+
+# Negation markers — even if a positive pattern also matches, presence
+# of any of these phrases means the action is NOT a positive capability
+# (e.g. ``"user cannot play"`` or ``"user can never log in"``).
+_NEGATION_PATTERNS: tuple["re.Pattern[str]", ...] = (
+    re.compile(r"\bcannot\b", re.IGNORECASE),
+    re.compile(r"\bcan ?not\b", re.IGNORECASE),
+    re.compile(r"\bcan't\b", re.IGNORECASE),
+    re.compile(r"\bunable\b", re.IGNORECASE),
+    re.compile(r"\bnever\b", re.IGNORECASE),
+    re.compile(r"\bwon't\b", re.IGNORECASE),
+    re.compile(r"\bwill not\b", re.IGNORECASE),
 )
 
 # Vague verbs / phrases that pass the "user can" check but carry no
@@ -82,8 +96,10 @@ class UserOutcome:
         coverage check.  Convention: ``outcome_<short_action_verb>``.
     action : str
         Concrete user capability statement.  Must contain a phrase
-        from :data:`_USER_CAPABILITY_PHRASES`.  Vague phrasings such as
-        "user can use the app" are rejected.
+        from :data:`_USER_CAPABILITY_PATTERNS` (word-boundary regex).
+        Vague phrasings such as "user can use the app" are rejected;
+        negations such as "user cannot ..." or "user can't ..." are
+        also rejected via :data:`_NEGATION_PATTERNS`.
     success_signal : str
         Observable evidence that the outcome is satisfied.  Drives the
         Integration Verification gate (Stage 5, deferred): a Playwright
@@ -108,7 +124,7 @@ class UserOutcome:
     scope: str
 
     def __post_init__(self) -> None:
-        """Validate all fields; reject vague or malformed outcomes."""
+        """Validate all fields; reject vague, malformed, or negated outcomes."""
         if not self.id:
             raise ValueError("UserOutcome.id must be a non-empty string")
 
@@ -124,13 +140,25 @@ class UserOutcome:
                 f"got {self.scope!r}"
             )
 
-        action_lc = self.action.lower()
-        if not any(phrase in action_lc for phrase in _USER_CAPABILITY_PHRASES):
+        # Reject negations FIRST — "user cannot play" contains "user can"
+        # as a substring, so word-boundary positive matching alone is not
+        # enough.  An action with a negation marker is never a positive
+        # capability regardless of whether a positive pattern also matches.
+        if any(p.search(self.action) for p in _NEGATION_PATTERNS):
             raise ValueError(
-                "UserOutcome.action must express a user capability "
-                f"(one of {list(_USER_CAPABILITY_PHRASES)}); got {self.action!r}"
+                "UserOutcome.action contains a negation (cannot / can't / "
+                "unable / never / won't / will not) — outcomes must "
+                f"express positive capabilities; got {self.action!r}"
             )
 
+        if not any(p.search(self.action) for p in _USER_CAPABILITY_PATTERNS):
+            raise ValueError(
+                "UserOutcome.action must express a user capability "
+                "(e.g. 'user can <verb>', 'users are able to <verb>'); "
+                f"got {self.action!r}"
+            )
+
+        action_lc = self.action.lower()
         if any(vague in action_lc for vague in _VAGUE_PHRASES):
             raise ValueError(
                 f"UserOutcome.action is too vague to be verifiable: "
@@ -248,6 +276,20 @@ async def extract_user_outcomes(
                 f"User-outcome extractor: outcome at index {idx} is not "
                 f"an object: {item!r}"
             )
+
+        # Reject null / non-string fields rather than coerce.  str(None)
+        # is the string "None" which is non-empty and would silently
+        # bypass UserOutcome's empty-string checks, polluting outcome ids
+        # used by compute_coverage.
+        for required_field in ("id", "action", "success_signal", "scope"):
+            value = item.get(required_field)
+            if not isinstance(value, str):
+                raise ValueError(
+                    f"User-outcome extractor: outcome at index {idx} "
+                    f"field {required_field!r} must be a string, got "
+                    f"{type(value).__name__}: {value!r}"
+                )
+
         try:
             outcomes.append(
                 UserOutcome(
@@ -273,12 +315,3 @@ async def extract_user_outcomes(
         seen_ids.add(outcome.id)
 
     return outcomes
-
-
-# Re-exported for type-checking convenience; the regex is exposed in case
-# future callers need to test action strings without constructing an
-# :class:`UserOutcome` first.
-USER_CAPABILITY_PATTERN = re.compile(
-    "|".join(re.escape(p) for p in _USER_CAPABILITY_PHRASES),
-    flags=re.IGNORECASE,
-)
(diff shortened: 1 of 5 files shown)


lwgray:
Addressed all three of @drchen's review concerns in commit 62bd7d1c:

### Concern 1: Default keyword mapper was a footgun

**Fix**: `compute_coverage`'s `mapper` parameter is now **required** — no implicit default. `_default_mapper` is renamed to `keyword_overlap_mapper` (public, exported) with a prominent warning docstring documenting the v31 false-positive risk.

A new `test_mapper_argument_is_required` locks in the guard: omitting the mapper raises `TypeError`.

For production use, a new `compute_coverage_with_llm` issues a single LLM call that maps all outcomes to all tasks at once. The prompt explicitly distinguishes "addresses" (produces user-visible evidence) from "supports" (internal logic), and uses the snake_game-v31 case as the example. This is what the decomposer integration will call.

### Concern 2: Vague-phrase blocklist was a band-aid pattern

**Fix**: Added a `DO NOT EXPAND` warning block above `_VAGUE_PHRASES` in `outcome_extractor.py`. Documents the structural fix (LLM specificity self-check) so the next contributor adds the self-check rather than extending the list — preventing the accumulation pattern that produced `frontend`/`e-commerce`/`UI-UX` band-aids in `advanced_parser.py`.

### Concern 3: Integration test hardcoded the answer

**Fix**: Removed `_visual_play_mapper` (the hardcoded-verb shortcut). The snake-v31 regression test now mocks two `compute_coverage_with_llm` calls (one before gap-fill returning empty coverage, one after returning the new rendering task as covering). The test now asserts pipeline behavior given honest coverage signals, not given a pre-encoded answer.

### Test impact

- 38/38 tests pass (was 30 — added 7 new for `compute_coverage_with_llm` plus the required-mapper guard)
- 990/990 in the full unit suite — no regressions
- mypy --strict / black / isort / flake8 / pydocstyle all clean

The footgun is now closed, the band-aid is documented as such, and the regression test asserts something meaningful about the pipeline rather than confirming a hand-crafted oracle.

lwgray:
Addressed all three Codex review issues in commit 65f09e04:

### Codex P1: Null coercion in extract_user_outcomes

**Fix**: Replaced `str(item.get(field, ""))` coercion with explicit `isinstance(value, str)` check before construction. Null and non-string values now raise `ValueError` naming the field and observed type.

Tests: `test_null_id_field_is_rejected`, `test_null_success_signal_is_rejected`, `test_non_string_field_is_rejected`.

### Codex P1: Null coercion in fill_gaps

**Fix**: Same pattern in `fill_gaps` — validate `isinstance` before `.strip()`. No more silent `str(None) → "None"` slipping through the empty-name check.

Tests: `test_null_name_field_is_rejected`, `test_null_description_field_is_rejected`, `test_non_string_name_field_is_rejected`.

### Codex P2: Substring matching accepts negations

**Fix**: Replaced `_USER_CAPABILITY_PHRASES` (substring tuple) with `_USER_CAPABILITY_PATTERNS` (compiled regex with `\b` word boundaries and a `(?!')` lookahead to reject `can't` contractions). Added `_NEGATION_PATTERNS` (cannot, can not, can't, unable, never, won't, will not). Negation check runs first — presence of any negation marker invalidates the outcome regardless of whether a positive pattern also matches.

Tests: `test_negation_cannot_is_rejected`, `test_negation_cant_contraction_is_rejected`, `test_negation_unable_is_rejected`, `test_negation_never_is_rejected`, `test_capability_phrase_must_match_at_word_boundary`.

### Test impact

- 49/49 tests pass (was 38, added 11 new for these regressions)
- 1001/1001 in full unit suite
- mypy --strict / black / isort / flake8 / pydocstyle clean

All three are now genuine schema violations rather than silently-coerced placeholders.

lwgray:
**Refactor commit `af9a70c3`** — replaced the three denylists (`_USER_CAPABILITY_PATTERNS`, `_NEGATION_PATTERNS`, `_VAGUE_PHRASES`) with a single LLM self-check pass.

The user pointed out that we'd accumulated exactly the band-aid pattern this whole module was built to escape — pattern matching against known-bad strings to do semantic work. Codex's substring-negation catch was a symptom of that growth.

### Structural change

`UserOutcome.__post_init__` is now SHAPE-only:
- id non-empty
- action non-empty (after stripping)
- success_signal non-empty
- scope in `{in_scope, out_of_scope}`

Semantic correctness is enforced by a new `filter_to_verifiable_capabilities(outcomes, llm_client)`. One LLM call asks "is this a concrete positive verifiable user capability?" for every outcome and returns YES/NO verdicts. Outcomes marked NO are dropped and logged at INFO level. `extract_user_outcomes` now calls this filter at the end of its pipeline.

### Why this beats denylists

The LLM understands every phrasing we'd otherwise need to enumerate — `"user is forbidden from"`, `"user shall avoid"`, `"user has zero ability to"`, the next antipattern we haven't seen yet — without us maintaining a list. Same pattern as `compute_coverage_with_llm` replacing keyword overlap.

### Test changes

- Removed dataclass tests for capability/negation/vague phrases (no longer construction invariants)
- Added `test_dataclass_does_not_enforce_semantic_constraints` to lock in the design point
- Renamed `test_vague_outcomes_are_rejected_at_extraction` → `test_vague_outcomes_are_rejected_by_llm_filter` (more accurate)
- Added `test_negation_outcomes_are_rejected_by_llm_filter`
- New `TestFilterToVerifiableCapabilities` class — 8 tests covering keep, drop, empty input (no LLM call), missing verdict default-to-keep, malformed JSON, missing key, non-boolean verifiable, and end-to-end filter-rejects-everything

### Numbers

- 53/53 tests pass (was 49)
- 1005/1005 in full unit suite
- All linters/typecheckers clean

The audit trail (out-of-scope outcomes preserved) is unchanged. Cost trade-off: one additional LLM call per project at planning time — acceptable; replaces the maintenance burden of growing denylists.

lwgray:
**Commit `82df4db8`** — removed all bad-case enumerations from both prompts. Following Kaia's review of the previous refactor, which correctly noted the denylist had moved into the filter prompt rather than disappearing.

### Architectural change

Both prompts now reference a single shared constant `_USER_OUTCOME_CRITERIA` that defines an outcome by three positive properties:

1. **CONCRETE** — names specific user-visible behavior such that an observer running the product can point at what they see, hear, or otherwise sense.
2. **POSITIVE** — names a capability the user gains, framed as something they CAN do. An absence, restriction, or prohibition is not a capability.
3. **VERIFIABLE** — its satisfaction is decidable from observing the running product. Evidence either appears or it does not.

**No enumerated bad-case phrases anywhere.** No `cannot/forbidden/unable` word lists. No `use the app/do stuff/interact with it` example denylists. No "Reject outcomes that are: Negations / Vague / Non-user-facing / Spec restatement" category bullets. The criteria define what an outcome MUST be; the LLM evaluates against the properties and generalizes to phrasings we never enumerated.

### What was removed

From `_FILTER_PROMPT`:
- Four-bullet "Reject outcomes that are any of:" enumeration
- All concrete bad-phrasing examples

From `_EXTRACTION_PROMPT`:
- "Every action must contain 'user can' (or 'users can' / 'is able to')" — pattern-imposing rule
- "'user can use the app' is too vague — reject this phrasing" — denylist example

What remains in the extraction prompt: ONE positive canonical example (`"user can play the snake game in their browser"`) for LLM grounding, plus output-shape rules (out_of_scope handling, non-empty list, success_signal requirement) — those are schema requirements, not denylists.

### Verification

```python
$ python -c "<grep for known bad phrases>"
CLEAN: no enumerated bad-case phrases in any prompt
```

(One generic-English `never` remains in "never return an empty list" — directive about output behavior, not a denylist phrase.)

### Tests

53/53 #449 tests pass; 1005/1005 in full unit suite. Test mocks check response handling, not prompt text, so no test changes were required — the API surface is unchanged.

### Why this is the right shape now

Compare:
- **Before:** prompt listed "use the app", "interact with it", "do stuff" → LLM matches what's listed
- **After:** prompt defines CONCRETE → LLM generalizes to any non-concrete phrasing including ones we never thought of

Properties generalize. Lists do not.

Issue #456, opened 2026-04-29
refactor: unify spec_coverage and outcome_coverage behind GraphAugmenter interface (fixes orphan synthesized tasks)


lwgray:
## Problem

Two parallel gap-detection pipelines exist with overlapping scope and divergent integration:

1. **`spec_coverage`** (`src/integrations/spec_coverage.py`) — extracts feature names from the spec, synthesizes `Implement {feature.title()}` tasks for missing ones.
2. **`outcome_coverage`** (#449, `src/marcus_mcp/coordinator/outcome_coverage.py`) — extracts user outcomes from the spec, synthesizes gap-fill tasks for missing ones.

The redundancy is one issue. The bigger one is that `spec_coverage` runs in the wrong place in the pipeline, producing **orphan synthesized tasks** that bypass dependency inference and foundation wiring.

## Symptom

Visible in v37 snake-game DAG: `Implement Speed Progression` appears as a top-level orphan node with no incoming edges, pointing only at `Integration verification`. Expected: it should depend on `Tech Foundation Setup` or `Game State Data Structure`.

Bottlenecks (orange) appear on the foundation tasks but the synthesized impl-task floats above them with no parents.

## Root Cause

`check_spec_coverage` is called in `nlp_tools.py:1251` (inside `create_project_from_description`) **after**:
- `parse_prd_to_tasks` (which runs `_infer_smart_dependencies`)
- Foundation wiring at `nlp_tools.py:332` (which makes domain tasks depend on foundation)
- `apply_safety_checks`

Synthesized tasks are constructed with `dependencies=[]` (`spec_coverage.py:235`) and never pick up:
- Foundation task dependencies (wiring already happened)
- Pattern-based dependencies (inference already happened)
- PRD-bundled design dependencies (also already happened)

Their only downstream connection is `enhance_project_with_integration`, which adds them to integration's dep list. So they become parents of the integration task with no parents themselves — orphans.

By contrast, `outcome_coverage` (#449) runs **inside** `parse_prd_to_tasks` between `_create_detailed_tasks` and `_infer_smart_dependencies`, so its synthesized tasks fully participate in dependency inference and foundation wiring.

## Secondary Issue: Naming Opacity

[...]

PR #466, opened 2026-04-30
fix(#455): add setup tier — latent classifier bug (NOT v37 orphan fix; that's #456)

### LINKED ISSUE #455: fix: classifier rejects setup tasks as prereqs of impl tasks via pattern matching (latent bug)
## Problem

``_is_logical_dependency`` in ``src/intelligence/dependency_inferer.py`` uses a four-tier task ordering (``design``<``implementation``<``testing``<``deployment``) to prevent circular dependencies, but tasks that don't match any of the four classifier patterns fall to ``other`` with order **2.5** — between implementation (2.0) and testing (3.0).

Result: foundation/setup tasks like ``Tech Foundation Setup`` (no ``design``/``implement``/``test``/``deploy`` keywords) are classified as ``other`` and rejected as prerequisites of implementation tasks **via the pattern-matching code path**.

## Scope clarification (added 2026-04-30)

**This is a latent bug, not the v37 user-visible orphan symptom.**  Investigation in kanban.db showed v37's ``Implement Speed Progression`` was orphaned because it was a **spec_coverage gap task** with empty ``dependencies=[]`` (description: "Spec coverage check found this feature missing from the task list..."), NOT because the classifier rejected its parent edge.  Tech Foundation Setup ``does`` have outbound edges to its sibling impl tasks (verified via ``task_dependencies`` rows) — those edges come from explicit foundation wiring at ``nlp_tools.py:332``, which bypasses the classifier entirely.

The actual v37 fix is **#456** (GraphAugmenter unification — folds spec_coverage into pre-inference pipeline).

This issue (#455) fixes a separate, latent correctness bug: the classifier's logical-order gate would reject a setup→impl dep ``if`` the pattern-matching path produced one.  In current production code the explicit foundation wiring in ``nlp_tools.py`` adds the edge regardless, so the classifier rejection is dead-letter.  But the rejection ``would`` matter for:

- AI-inferred deps in ``HybridDependencyInferer`` ambiguous paths
- Pattern matches between setup-named tasks not covered by the explicit foundation wiring (e.g., a hypothetical ``Initial Database Migration`` synthesized by future code)
- Any future code path that doesn't go through the explicit foundation wiring

## Reproduction (latent path only)

``Implement Speed Progression`` (classified ``implementation``, order 2) attempting to depend on ``Tech Foundation Setup`` (classified ``other``, order 2.5) **via the pattern-matching path**:

```python
# dependency_inferer.py:399
if dependency_order >= dependent_order:
    return False  # 2.5 >= 2.0 → dependency rejected
```

``setup_blocks_all`` pattern matches the regex (``Tech Foundation Setup`` contains ``setup``, ``Implement Speed Progression`` contains ``implement``), but the logical-order check then rejects it.

In production code the same edge is added by ``nlp_tools.py:332`` foundation wiring, so users don't see the rejection.  But the classifier is still wrong.

## Root Cause

``_classify_task_type`` keyword set is too narrow:

```python
# dependency_inferer.py:454
if any(word in name_lower for word in ["implement", "build", "create", "develop", "code", "write"]):
    return "implementation"
```

``Tech Foundation Setup`` has none of these.  Same for any LLM-generated foundation/setup task that doesn't happen to contain an implementation verb.

## Proposed Fix

Add a ``setup`` tier (order 0.5, below design) with explicit keywords (``setup``, ``init``, ``initial``, ``configure``, ``install``, ``scaffold``, ``foundation``).  Tier-checking order in ``_classify_task_type`` puts setup last so semantically specific tiers win when keywords overlap (``"Design Initial Setup"`` stays a design task).

## Acceptance Criteria

- [x] Setup-named foundation tasks (``Tech Foundation Setup``, ``Install Dependencies``, ``Initialize Database``) classify as ``setup``, not ``other``
- [x] Setup tasks (order 0.5) can pass the logical-order gate as prereqs of design / impl / test / deploy
- [x] No regression: design → impl, impl → test, etc. still work
- [x] Same-tier edges (setup → setup) still rejected to prevent cycles
- [x] PR body documents this is a latent fix, not the v37 user-visible orphan fix

## Out of Scope

- ``Game State Data Structure`` → ``Implement X``: no current pattern matches that pair.  Different bug, separate fix.  Not blocked by this issue.
- The v37 user-visible orphan (``Implement Speed Progression``) is a spec_coverage bug, fixed by #456.

lwgray:
Closes #455.  v0.3.6.post2 milestone.

## Scope correction (added 2026-04-30)

This PR fixes a **latent** classifier bug, NOT the v37 user-visible orphan symptom.  Original PR title implied this was the v37 fix; that was wrong.  Investigation in kanban.db showed v37's ``Implement Speed Progression`` was orphaned because it was a **spec_coverage gap task** with empty ``dependencies=[]``, NOT because the classifier rejected its parent edge.  The actual v37 fix is **#456** (GraphAugmenter unification — folds spec_coverage into pre-inference pipeline so synthesized tasks participate in dependency inference + foundation wiring).

What this PR DOES fix: the classifier's logical-order gate would reject ``setup → impl`` deps via the pattern-matching code path.  In current production, explicit foundation wiring at ``nlp_tools.py:332`` adds the edge regardless, so the rejection is dead-letter.  But the classifier is still wrong, and the rejection would matter for:

- AI-inferred deps in ``HybridDependencyInferer`` ambiguous paths
- Pattern matches between setup-named tasks not covered by the explicit foundation wiring
- Any future code path that doesn't go through the explicit foundation wiring

Cheap defensive correctness improvement.  Small scope, well-tested.

## Summary

- Adds ``"setup"`` tier (order 0.5) to dependency classifier so foundation/scaffolding tasks (``Tech Foundation Setup``, ``Install Dependencies``, ``Initialize Database``) can be valid prerequisites of design and implementation tasks via pattern matching.
- Pre-fix: setup-named tasks classified as ``"other"`` (order 2.5) and the logical-order gate rejected them as prereqs of impl (order 2.0).

## Tier-precedence design

Setup is checked LAST among named tiers in ``_classify_task_type``.  This keeps semantically specific tiers winning when keywords overlap:

- ``"Design Initial Setup Approach"`` → design (not setup)
- ``"Implement Initial Configuration"`` → implementation (not setup)
- ``"Tech Foundation Setup"`` → setup ✓ (no other tier matches)

Setup tier is below design in ``task_order`` (0.5 vs 1.0), so even when a future pattern fires setup→design, the order check accepts.

## Out of scope

``Game State Data Structure`` → ``Implement X``: no current dependency pattern fires on that pair.  Different bug, separate fix.  Should be filed separately if it surfaces as a real production gap.

The v37 user-visible orphan: see #456.  This PR does NOT fix it.

## Test plan

- [x] ``pytest -m unit`` — 1103 passed (+16 new in this PR)
- [x] ``mypy --strict`` clean on modified files
- [x] ``black`` / ``isort`` / ``flake8`` clean
- [x] Pre-commit hooks pass
- [x] Kaia review applied: two tests strengthened to invoke ``_is_logical_dependency`` directly instead of trusting the order math (commit ``0686a3a3``)

## Files

- ``src/intelligence/dependency_inferer.py`` — setup tier in ``task_order``, setup keywords in ``_classify_task_type``
- ``tests/unit/intelligence/test_classifier_setup_tier.py`` — 16 tests covering classification, tier precedence, logical-dep gate (setup→impl positives, design→impl + impl→test no-regression checks, same-tier cycle prevention)

## Related

- #456 — GraphAugmenter unification (the actual v37 ``Implement Speed Progression`` fix)
- #466 (this PR) — separate latent classifier correctness improvement

### DIFF
--- a/src/intelligence/dependency_inferer.py
+++ b/src/intelligence/dependency_inferer.py
@@ -15,6 +15,25 @@ from src.core.models import Task, TaskStatus
 
 logger = logging.getLogger(__name__)
 
+# Setup-tier keyword matcher (issue #455 / Codex P2 on PR #466).
+# Word-boundary anchored prefix match.  Prefixes are intentionally
+# shorter than the canonical word so common inflections match:
+#
+#   ``\binit\w*``     → init, initial, initialize, initialization
+#   ``\bconfig\w*``   → config, configure, configured, configuration
+#   ``\binstall\w*``  → install, installs, installer, installation
+#   ``\bscaffold\w*`` → scaffold, scaffolds, scaffolding
+#   ``\bsetup\w*``    → setup, setups
+#   ``\bfoundation\w*`` → foundation, foundations, foundational
+#
+# The ``\b`` boundary rejects un-/re-/pre-/post-prefixed false
+# positives: ``uninstall`` (preceded by 'n' — word char, no boundary),
+# ``reconfigure`` (preceded by 'e'), ``presetup``, ``postinit``.
+_SETUP_PATTERN: re.Pattern[str] = re.compile(
+    r"\b(setup|init|config|install|scaffold|foundation)\w*",
+    re.IGNORECASE,
+)
+
 
 @dataclass
 class DependencyPattern:
@@ -488,18 +507,24 @@ class DependencyInferer:
         # among named tiers so design / impl / testing / deployment
         # win when keywords overlap (e.g. "Design initial setup"
         # stays a design task).
-        if any(
-            word in name_lower
-            for word in [
-                "setup",
-                "init",
-                "initial",
-                "configure",
-                "install",
-                "scaffold",
-                "foundation",
-            ]
-        ):
+        #
+        # Word-boundary matching (Codex P2 / Claude bot review on
+        # PR #466): plain ``substring in name`` would match
+        # ``"install"`` inside ``"uninstall"``, classifying teardown
+        # tasks as setup and (combined with the new setup order 0.5)
+        # making them eligible prerequisites of impl tasks via
+        # ``setup_blocks_all`` — a regression my own PR would
+        # introduce.  Anchored prefix match (``\bword\w*``) accepts:
+        #
+        # - ``\binit\w*``      → init, initial, initialize, initialization
+        # - ``\binstall\w*``   → install, installs, installation
+        # - ``\bconfigure\w*`` → configure, configured, configuring
+        #
+        # And rejects:
+        # - ``uninstall`` (preceded by 'n', no \b boundary at 'i')
+        # - ``reconfigure`` (preceded by 'e', no \b boundary)
+        # - ``presetup`` / ``postsetup`` (preceded by word chars)
+        if _SETUP_PATTERN.search(name_lower):
             return "setup"
 
         return "other"
--- a/src/intelligence/dependency_inferer.py
+++ b/src/intelligence/dependency_inferer.py
@@ -383,8 +383,17 @@ class DependencyInferer:
         dependent_task_type = self._classify_task_type(dependent_task.name)
         dependency_task_type = self._classify_task_type(dependency_task.name)
 
-        # Define logical task ordering: design < implementation < testing < deployment
+        # Define logical task ordering:
+        # setup < design < implementation < testing < deployment
+        #
+        # ``setup`` (issue #455) sits below design so foundation /
+        # scaffolding tasks (e.g. "Tech Foundation Setup", "Install
+        # Dependencies") are valid prerequisites of design and impl
+        # tasks.  Pre-#455 they classified as ``other`` (order 2.5)
+        # which blocked them from being prereqs of impl (order 2.0)
+        # and produced orphan top-level nodes in the DAG.
         task_order = {
+            "setup": 0.5,
             "design": 1,
             "implementation": 2,
             "testing": 3,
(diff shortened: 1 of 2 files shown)


lwgray:
## Codex P2 + Claude bot review addressed

**Both reviewers were right — this was an actual regression my PR introduced**, not just a hypothetical false positive.

Pre-PR: `"Uninstall Legacy Dependencies"` → `"other"` (2.5) → order gate REJECTS → impl (2.5 ≥ 2) ✓
Mid-PR: `"Uninstall Legacy Dependencies"` → `"setup"` (0.5) → order gate ACCEPTS → impl (0.5 < 2) ❌ **regression**
Now (commit `1d79059d`): word-boundary regex correctly classifies as `"other"` again.

### Fix details

Compiled regex at module level: `\\b(setup|init|config|install|scaffold|foundation)\\w*`

Prefixes intentionally shorter than canonical words so inflections match:
- `\\binit\\w*` → init, initial, initialize, initialization
- `\\bconfig\\w*` → config, configure, configured, configuration
- `\\binstall\\w*` → install, installs, installer, installation

Word boundary rejects un-/re-/pre-/post- prefixed false positives.

### New tests (10)

`TestClassifierWordBoundaryFalsePositives` covers 7 parametrized teardown/redo cases plus 3 inflection-still-works tests.

Total: 1113 unit tests pass (+26 in this PR, +10 from this commit).

Thanks for catching this — would have shipped a regression otherwise.

PR #473, opened 2026-05-01
refactor(#456): unify outcome_coverage + spec_coverage behind GraphAugmenter Protocol

### LINKED ISSUE #456: refactor: unify spec_coverage and outcome_coverage behind GraphAugmenter interface (fixes orphan synthesized tasks)

lwgray:
## Summary

Closes #456. Replaces the parallel `outcome_coverage` (in-decomposer) and `spec_coverage` (post-safety-check) pipelines with a single `GraphAugmenter` Protocol + chain orchestrator. Both augmenters now run inside the decomposer between `_create_detailed_tasks` and `_infer_smart_dependencies`, so synthesized gap tasks flow through dependency inference and foundation wiring as first-class graph members instead of being appended later as orphans.

**Headline fix:** v37 orphan-failure-mode (spec_coverage gap tasks landing with `dependencies=[]`) is closed.

## What changed

- **New `GraphAugmenter` Protocol** + `AugmentationResult` dataclass at `src/marcus_mcp/coordinator/graph_augmentation.py`.
- **New `run_augmenter_chain` orchestrator** with defense-in-depth `try/except` around each augmenter, augmenter-name uniqueness validation at chain entry, telemetry namespacing by `augmenter.name`.
- **`OutcomeCoverageAugmenter`** dispatches to two new public module functions in `outcome_coverage.py`: `apply_outcome_coverage_to_feature_graph` and `apply_outcome_coverage_to_contract_graph`. The legacy parser-side helpers (`_apply_outcome_coverage_to_*`, `_build_*_gap_fill_task`, `ParserOutcomeCoverage` dataclass) are deleted.
- **`SpecCoverageAugmenter`** wraps the existing `check_spec_coverage` function. The legacy post-safety-check call site at `nlp_tools.py:1336` is removed. The `project_name` parameter (declared but unused) is dropped.
- **`AdvancedPRDParser._build_augmenter_chain()`** is the single registration site; both decomposer paths route through it.
- **ADR 0011** documents the pattern and the recipe for adding future augmenters.

## Bright-line gates (all pass)

```
git grep _apply_outcome_coverage -- src/   = 0
git grep ParserOutcomeCoverage    -- src/   = 0
git grep "from src.integrations.spec_coverage import check_spec_coverage" -- src/ = 1
```

## Stages

This landed as a 5-stage TDD migration with a Kaia review gate after each stage:

- **Stage 1** — Protocol + dataclass (12 tests)
- **Stage 2** — `OutcomeCoverageAugmenter` wrapper around legacy helpers (17 tests)
- **Stage 3** — `run_augmenter_chain` orchestrator + decomposer cutover (15 chain tests)
- **Stage 4** — `SpecCoverageAugmenter` joins chain, legacy call site removed (17 tests)
- **Stage 5** — Lift helpers, delete legacy, drop `project_name`, factor chain helper, ADR

## Test plan

- [x] `pytest -m unit` — 1208 passed, 1 skipped, no regressions
- [x] `mypy --strict` clean on all touched src files
- [x] `black` + `isort` + `pydocstyle` + `flake8` + `bandit` all pass via pre-commit
- [x] Augmenter-name uniqueness check fails loud on duplicate registration
- [x] v32 regression-canary test (`test_snake_v32_intent_fidelity`) still green
- [x] Spec-gap integration-order test (`test_spec_gap_integration_order`) still green
- [x] PLANNING_INTENT_FIDELITY event payload byte-identical to pre-#456 shape
- [x] Bright-line PR gates verified

### DIFF
--- a/docs/architecture/adr/0011-graph-augmenter-protocol.md
+++ b/docs/architecture/adr/0011-graph-augmenter-protocol.md
@@ -238,6 +238,15 @@ Pattern for future augmenters (NFR coverage, security checks, etc.):
    registration time, but verify against existing augmenter names
    to avoid the `ValueError` at boot.
 
+6. **Stay stateless.** Augmenters are constructed per-call by
+   `AdvancedPRDParser._build_augmenter_chain()`. Adding instance
+   state (caches, retry budgets, rate-limit windows) silently
+   discards it across calls — see the lifetime-expectation note in
+   :class:`OutcomeCoverageAugmenter` for the tripwire pattern. If an
+   augmenter genuinely needs persistent state, audit the
+   construction site and move it to per-decomposer-instance or
+   per-process scope before adding the attribute.
+
 ---
 
 ## References
--- a/docs/architecture/adr/0011-graph-augmenter-protocol.md
+++ b/docs/architecture/adr/0011-graph-augmenter-protocol.md
@@ -1,0 +1,258 @@
+# ADR 0011: GraphAugmenter Protocol for Pre-Inference Task Synthesis
+
+**Status:** Accepted
+
+**Date:** 2026-04-30
+
+**Deciders:** Marcus Core Team
+
+**Issue:** [#456](https://github.com/lwgray/marcus/issues/456)
+
+---
+
+## Context
+
+Pre-#456, Marcus had two parallel pipelines that synthesized tasks
+into the graph after the decomposer produced its initial task list:
+
+1. **Outcome coverage** (`outcome_coverage` — issue #449) ran *inside*
+   the decomposer between `_create_detailed_tasks` and
+   `_infer_smart_dependencies`. Synthesized gap-fill tasks correctly
+   participated in dependency inference and foundation wiring.
+
+2. **Spec coverage** (`spec_coverage`) ran *after* the decomposer +
+   safety checks at `nlp_tools.py:1336`. Synthesized gap tasks were
+   appended to the task list with `dependencies=[]` — orphans that
+   bypassed both `_infer_smart_dependencies` and foundation wiring.
+   This was the v37 orphan-failure-mode.
+
+Both pipelines did the same fundamental thing: "look at the graph,
+detect a gap, synthesize a task to fill it." But they integrated at
+different points in the pipeline with different failure modes.
+
+The forcing function: a third would-be augmenter (NFR coverage,
+security checks, lint coverage…) faced an unclear question — *where*
+in the pipeline does it slot in? Each new addition risked being
+either (a) another orphan, like spec_coverage, or (b) a parallel
+implementation of the same lift-and-fill pattern.
+
+### Requirements
+
+- Single integration point for pre-inference task synthesis so all
+  augmenters land at the same correct location.
+- Synthesized tasks always participate in dependency inference and
+  foundation wiring as first-class graph members.
+- Defense-in-depth against augmenter exceptions — one buggy augmenter
+  must never crash the decomposer.
+- Telemetry namespace per augmenter so future additions don't
+  collide on event-payload keys.
+- Compatible with Marcus's existing duck-typing convention (no
+  inheritance-from-base requirement).
+- Observable: failures must be loud and named in logs so operators
+  can diagnose which augmenter failed without grepping for stack
+  traces.
+
+---
+
+## Decision
+
+Define a structural Protocol — `GraphAugmenter` — and a chain
+orchestrator — `run_augmenter_chain` — both in
+`src/marcus_mcp/coordinator/graph_augmentation.py`. Both decomposers
+(`parse_prd_to_tasks` and `decompose_by_contract`) route their
+post-`_create_detailed_tasks` task list through the chain before
+calling `_infer_smart_dependencies`.
+
+### The Protocol
+
+```python
+@runtime_checkable
+class GraphAugmenter(Protocol):
+    name: str
+
+    async def augment(
+        self,
+        *,
+        prd_analysis: "PRDAnalysis",
+        tasks: List[Task],
+        contract_artifacts: Optional[Dict[str, Any]] = None,
+    ) -> AugmentationResult: ...
+```
+
+**Why a Protocol (structural) and not an ABC (nominal):** Marcus's
+existing pipeline conventions favor duck typing; structural typing
+keeps the interface non-invasive to existing classes.
+`@runtime_checkable` lets the chain orchestrator validate registered
+augmenters with `isinstance` when registration happens.
+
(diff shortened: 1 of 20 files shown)


lwgray:
## Codex P2 addressed in `31e23961`

Confirmed the regression. Pre-Stage 4, `check_spec_coverage` ran post-safety-check on the combined `ghost + foundation + contract` task list. After Stage 4 the augmenter chain saw only contract tasks, making foundation tasks invisible to spec_coverage's keyword scan and creating false-positive duplicate gap tasks.

**Fix:** thread `foundation_tasks` into `decompose_by_contract` via a new `pre_existing_tasks: Optional[List[Task]] = None` parameter. The chain now runs on the combined list so augmenters see the full pre-fork graph during scanning. Orchestrator's final assembly drops the redundant `+ foundation_tasks` concat since they're already in `augmented_tasks`.

**Side effect cleanup:** the foundation-task dep-wiring loop now skips foundation tasks themselves (they're in the iteration target now and would otherwise add `foundation_id` as their own dependency).

**Regression tests added:**
- `test_pre_existing_tasks_visible_to_chain` — spies on contract-graph helper to confirm foundation tasks appear in chain input
- `test_pre_existing_tasks_default_none_backward_compat` — ensures backward-compat without the new param

**Acknowledged limitation:** ghost tasks (DONE-status design tasks) are still added by the orchestrator post-chain. They're unlikely to false-positive on spec keyword matching, and threading them would require a larger restructure. Documented inline.

1211 unit tests green (1209 + 2 new). Mypy strict clean.

PR #585, opened 2026-05-20
refactor(experiments): extract per-CLI harness Protocol + registry

lwgray:
## Why

`AgentSpawner` previously dispatched harness-specific behavior through **seven** `if self.harness == "claude" / elif self.harness == "codex"` sites spread across one ~2000-line file:

- `_build_mcp_register_snippet` — different `mcp add` syntax per CLI
- `_build_agent_command` — different CLI flags per harness
- `_build_worker_invocation_block` — wrapper loop (codex) or bare command (claude)
- `confirm_trust_if_prompted` — claude only
- pre-flight `which <cli>` — different binary name + install hint
- `_pretrust_directory` — claude only

Adding a third harness (`gemini` — next PR) meant editing every one of those sites and hoping nothing was missed. This PR collects every harness-specific behavior behind a single `Harness` Protocol with one implementation class per CLI. Adding a harness now means writing a new class; `spawn_agents.py` does not change.

**No behavior change.** Every rendered shell script is byte-identical before vs after — the existing `TestCodexScriptsRender` + sibling tests pass unchanged, which is the safety net.

> **About this project, briefly:** Marcus is a multi-agent system. The experiment runner spawns N agent CLIs (each in its own tmux pane) that coordinate via a shared Marcus MCP server. Today supported CLIs are `claude` (Anthropic) and `codex` (OpenAI). A "harness" in this PR's vocabulary is the per-CLI strategy object describing how the runner talks to that CLI.

### New file: `dev-tools/experiments/runners/harness.py`

A `Harness` Protocol (runtime-checkable) capturing the per-CLI surface the runner needs to know about, plus two implementation classes and a registry:

```python
class Harness(Protocol):
    name: str                              # "claude" | "codex"
    binary: str                            # PATH-lookup name
    provider: str                          # "anthropic" | "openai"  (for future cost-ingester scoping)
    needs_trust_dialog_poll: bool          # claude only
    needs_pretrust_directory: bool         # claude only
    workflow_files: tuple[str, ...]        # ("CLAUDE.md", "AGENTS.md")
    def build_agent_command(self, workdir, prompt_file, *, model_flag, print_mode) -> str: ...
    def wrap_worker_invocation(self, inner_cmd: str) -> str: ...
    def build_mcp_register_snippet(self) -> str: ...
    def install_hint(self, marcus_mcp_url: str) -> list[str]: ...

HARNESSES: dict[str, Harness] = {"claude": ClaudeHarness(), "codex": CodexHarness()}
def get_harness(name: str) -> Harness: ...
```

Both implementations carry the exact shell strings the prior `if/elif` branches produced. The long flag-by-flag explanatory comments that used to live in `_build_agent_command`'s if-branches now live inside the harness classes' docstrings, next to the code they explain.

### Modified: `dev-tools/experiments/runners/spawn_agents.py`

- Added a dual-path import that works under both load paths:
  - `from runners.spawn_agents import ...` (production, via `run_experiment.py`)
  - `importlib.util.spec_from_file_location(...)` (unit tests, no `runners` package on `sys.path`)
- New `AgentSpawner.harness_impl` **lazy property** — the lookup runs on first read, not in `__init__`. This keeps existing test fixtures that build a spawner from a `MagicMock` config working: only tests that actually exercise harness-specific behavior need to supply a real harness string.
- Replaced 7 dispatch sites with `self.harness_impl.<method>` calls. Net: ~119 lines removed, ~93 lines added (the deletions are mostly the long comment blocks that moved into the harness classes).
- `harness` argument validation against the registry's keys instead of an inline `{"claude", "codex"}` literal.

### New file: `tests/unit/experiments/test_harness.py`

24 unit tests pinning the Protocol contract directly:

- `TestRegistry` — name lookup, singleton, unknown-name `ValueError` lists supported choices
- `TestClaudeHarnessMetadata` / `TestCodexHarnessMetadata` — identity, trust-dialog flag, pretrust flag, workflow files
- `TestClaudeRenderedShell` / `TestCodexRenderedShell` — exact assertions on what the strings contain (mcp register snippet, agent command, worker wrapper loop including the `rc=0`-break from PR #554's Codex review, install hint)
- `TestProtocolConformance` — both implementations satisfy `isinstance(x, Harness)` via `@runtime_checkable`

The existing `test_spawn_agents.py` suite (134 tests) continues to cover the integration with `AgentSpawner`.

## Worked example — adding gemini after this PR

```python
# dev-tools/experiments/runners/harness.py
class GeminiHarness:
    name = "gemini"
    binary = "gemini"
    provider = "google"
    needs_trust_dialog_poll = False
    needs_pretrust_directory = False
    workflow_files = ("CLAUDE.md", "AGENTS.md", "GEMINI.md")

    def build_agent_command(self, workdir, prompt_file, *, model_flag, print_mode): ...
    def wrap_worker_invocation(self, inner_cmd): ...  # tbd — gemini -p exits per prompt
    def build_mcp_register_snippet(self): ...
    def install_hint(self, marcus_mcp_url): ...

HARNESSES["gemini"] = GeminiHarness()
```

That's the entire diff for harness support. `spawn_agents.py` does not change.

## Where to look in the code first

| File | Purpose |
|---|---|
| `dev-tools/experiments/runners/harness.py` | Protocol + `ClaudeHarness` + `CodexHarness` + `HARNESSES` registry + `get_harness` |
| `dev-tools/experiments/runners/spawn_agents.py` lines 22-50 | Dual-path import (package vs importlib) |
| `dev-tools/experiments/runners/spawn_agents.py` `harness_impl` property | Lazy lookup for bypass-init test fixtures |
| `tests/unit/experiments/test_harness.py` | New Protocol test suite (24 tests) |
| `tests/unit/experiments/test_spawn_agents.py` | Existing 134 tests — unchanged, all pass |

## Glossary

| Term | Meaning |
|---|---|
| **harness** | The per-CLI strategy object describing how the runner talks to one agent CLI |
| **`runtime_checkable` Protocol** | Python typing construct: structural typing (any class with the right methods/attrs "is" the Protocol) plus `isinstance()` works at runtime |
| **lazy property** | `@property` that resolves on first read and caches the result — avoids running expensive code at `__init__` |
| **dispatch site** | A place in the code that branches on harness name to do something different |
| **byte-identical** | The bytes of the rendered shell scripts compare equal before vs after this PR |

## How to verify it works

1. **No behavior change** — every existing test passes:

   ```bash
   pytest tests/unit/experiments/ -m unit -q
   ```

   Expected: `158 passed` (134 pre-existing + 24 new Protocol tests).

2. **mypy clean** on both files:

   ```bash
   mypy dev-tools/experiments/runners/spawn_agents.py \
        dev-tools/experiments/runners/harness.py \
        --config-file=pyproject.toml
   ```

   Expected: `Success: no issues found in 2 source files`.

3. **Rendered scripts identical** — the existing `TestCodexScriptsRender` and `TestClaudeModelFlagLandsInGeneratedScripts` classes pin the exact substring layout. They pass unchanged.

4. **Live smoke**: a codex run on the experiment runner should behave identically to develop's behavior (rendered shell scripts are the contract).

## Test plan

- [ ] `pytest tests/unit/experiments/ -m unit -q` → 158 passed
- [ ] `mypy dev-tools/experiments/runners/{spawn_agents,harness}.py` → clean
- [ ] Spot-check rendered project-creator + worker + monitor scripts against the develop versions — diff should be zero
- [ ] Adding a `GeminiHarness` (next PR) requires no edits to `spawn_agents.py`

## Related

- PR #554 — the codex harness landed; surfaced the need for this refactor
- Next: PR adding `GeminiHarness` against this registry (no `spawn_agents.py` edits)
- Issue #582 — codex cost-tracking parity (will use `harness_impl.provider`)

### DIFF
--- a/dev-tools/experiments/runners/harness.py
+++ b/dev-tools/experiments/runners/harness.py
@@ -1,0 +1,344 @@
+"""
+Per-CLI harness definitions for the Marcus experiment runner.
+
+A *harness* is the agent CLI a spawned tmux pane runs — currently
+``claude`` (Anthropic's Claude Code) or ``codex`` (OpenAI's Codex
+CLI).  Both speak the same Marcus MCP protocol, but the surface
+around them — the exact shell command, the worker-keepalive
+wrapper, the trust-dialog handling, the workflow-file convention,
+and the install hint shown on a missing binary — differs.
+
+Before this module landed, ``AgentSpawner`` had six ``if/elif`` sites
+spread across one ~2000-line file, one per harness-specific quirk.
+Adding a third harness (e.g. ``gemini``) meant editing every site.
+This module collects every harness-specific behavior into a single
+:class:`Harness` Protocol with one implementation per CLI, plus a
+:data:`HARNESSES` registry the spawner looks up by name.
+
+The contract is "byte-identical rendered shell scripts": the strings
+returned here must be exactly what the previous ``if/elif`` blocks
+returned, so the existing ``TestCodexScriptsRender`` and sibling
+tests stay green without modification.
+
+Classes
+-------
+Harness
+    Structural protocol every harness implementation must satisfy.
+ClaudeHarness
+    Default harness — Anthropic Claude Code CLI.
+CodexHarness
+    OpenAI Codex CLI harness with ``--enable goals`` + a bounded
+    bash relaunch loop.
+
+Functions
+---------
+get_harness(name)
+    Resolve a harness by name, raising :class:`ValueError` for
+    unknown names with a list of supported choices.
+
+Data
+----
+HARNESSES
+    Mapping of harness name -> :class:`Harness` instance.
+"""
+
+from pathlib import Path
+from typing import Dict, List, Protocol, Tuple, runtime_checkable
+
+
+@runtime_checkable
+class Harness(Protocol):
+    """Structural protocol every per-CLI harness implementation satisfies.
+
+    Attributes
+    ----------
+    name : str
+        Stable identifier the runner stores on ``config.harness`` and
+        the user picks on the command line (``--harness <name>``).
+    binary : str
+        Executable name looked up on PATH during the spawner
+        pre-flight.
+    provider : str
+        Cost-tracking provider label (``"anthropic"`` / ``"openai"``).
+        Used by future cost ingesters to scope ``token_events`` rows.
+    needs_trust_dialog_poll : bool
+        ``True`` when the harness raises an interactive trust dialog
+        on first launch in a directory; the spawner then polls each
+        pane for the prompt and confirms it.  Claude only.
+    needs_pretrust_directory : bool
+        ``True`` when the harness reads a per-directory trust file
+        (e.g. ``~/.claude.json``) that the spawner must seed before
+        launch.  Claude only.
+    workflow_files : Tuple[str, ...]
+        Workflow filenames the spawner copies into the implementation
+        directory so the agent picks up Marcus-specific guidance.
+        For codex, both ``CLAUDE.md`` and ``AGENTS.md`` are written —
+        the latter is the codex-specific convention
+        (https://developers.openai.com/codex/guides/agents-md).
+    """
+
+    name: str
+    binary: str
+    provider: str
+    needs_trust_dialog_poll: bool
+    needs_pretrust_directory: bool
+    workflow_files: Tuple[str, ...]
+
+    def build_agent_command(
+        self,
+        workdir: Path,
+        prompt_file: Path,
+        *,
+        model_flag: str,
+        print_mode: bool = False,
+    ) -> str:
+        """Return the per-pane shell command line that launches the agent."""
+        ...
+
+    def wrap_worker_invocation(self, inner_cmd: str) -> str:
+        """Return the worker shell block: bare invocation or wrapped loop."""
+        ...
+
+    def build_mcp_register_snippet(self) -> str:
+        """Return the snippet that registers Marcus MCP from inside a pane."""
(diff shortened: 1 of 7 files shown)


lwgray:
Codex P1 addressed (`44be73d0` → `cc7992ca`).

**The regression**
`spawn_agents.py` has an `if __name__ == "__main__":` block (line ~2143). After this PR's `from runners.harness import ...` import, direct invocation (`python dev-tools/experiments/runners/spawn_agents.py <dir>`) crashed with `ModuleNotFoundError: No module named 'runners'` — Python puts `runners/` on `sys.path`, not its parent.

**The fix**
Bootstrap the parent directory at module top, idempotent:

```python
_RUNNERS_PARENT = str(Path(__file__).resolve().parent.parent)
if _RUNNERS_PARENT not in sys.path:
    sys.path.insert(0, _RUNNERS_PARENT)

from runners.harness import HARNESSES, Harness, get_harness  # noqa: E402
```

Works under all three entry points (production via `run_experiment.py`, tests via conftest, direct script invocation).

**Regression test**
New `TestSpawnAgentsDirectInvocation` runs the script in a subprocess and asserts (1) no `ModuleNotFoundError` in stderr, (2) the usage line reaches stdout. Subprocess (not `importlib`) so the test mirrors the user-visible invocation — pytest's own `sys.path` doesn't mask the bug.

**Verified**
- `python dev-tools/experiments/runners/spawn_agents.py` → prints usage, no traceback
- `pytest tests/unit/experiments/ tests/unit/mcp/test_parallel_sqlite.py -q` → 174 passed
- mypy clean

